PR #1, opened 2022-05-13
Try adding CI using Github Actions

tcbrindle:


tcbrindle:
RIght, so the current Ubuntu runners only have GCC 11.1, which appears to fail because it doesn't have constexpr `optional::operator=` from C++20.

On the Windows side, everything compiles perfectly well, but on release builds we then fail a runtime test **on the exact same code that has just been `static_assert()`-ed during compilation**. I don't even know how that's possible.

I give up, for now at least.

PR #6, opened 2022-05-24
Add flux::to

tcbrindle:


tcbrindle:
Ignoring the weird Codecov report, I don't understand how the coverage goes down when more template code suddenly becomes "live"...

PR #12, opened 2022-10-11
Don't implicitly reference sequences in adaptors

tcbrindle:
Warning: huge, massive, invasive changes ahead.

Up until this point, we have basically followed the design of Ranges, albeit with "views" renamed to "lenses". Adaptors act on "lenses", and the factory functions automatically turn any sequence into a lens: for lvalues, by (effectively) by using flux::ref, and for rvalues by wrapping them in flux::owning_adaptor.

Unfortunately, this has a pretty major downside: adaptors very often end up implicitly holding references to sequences, with all the lifetime issues that entails. For example, saying:

    auto vec = std::vector{1, 2, 3, 4, 5};
    auto filter = flux::filter(vec, [](int i) { return true; });

Then `filter` contains a reference to `vec`, and its lifetime must not exceed that of the vector it is referencing: but this isn't made clear anywhere in the code. With these changes, you now need to explicitly say:

     auto filter = flux::filter(flux::ref(vec), [](int i) { return true; });

(or flux::ref(vec).filter(...)) which makes it clearer what is going on. This should also avoid various cases where we were uninitentionally wrapping a ref_view in an owning_view in a ref_view etc, which the optimiser can cut through okay but is less than ideal in debug builds (these were probably mistakes on my part, but still...)

With this commit, we modify the adaptable_sequence concept so that it is only satisified by *rvalue* sequences. This means that you must pass an rvalue to an adaptor function: this can be achieved either by making an explicit copy, an explicit move, or explicitly constructing a ref_adaptor via flux::ref (for lvalues). We're putting the user in charge, or at least that's the idea.

### DIFF
--- a/include/flux/core/concepts.hpp
+++ b/include/flux/core/concepts.hpp
@@ -286,8 +286,9 @@ constexpr bool is_ilist<T, std::initializer_list<E>> = true;
 template <typename Seq>
 concept adaptable_sequence =
     sequence<Seq> &&
-    (std::is_lvalue_reference_v<Seq> ||
-        (std::movable<std::remove_reference_t<Seq>> && !detail::is_ilist<Seq>));
+    !std::is_reference_v<Seq> &&
+    std::movable<Seq> &&
+    !detail::is_ilist<Seq>;
 
 template <typename D>
 struct lens_base;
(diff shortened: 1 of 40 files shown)


tcbrindle:
Codecov complaints seem to be a rounding error. MacOS builds are broken at the moment due to a linker bug in the latest Xcode toolset.

PR #13, opened 2022-11-14
Fix Github Actions again

tcbrindle:
...or try to, anyway


tcbrindle:
Mac OS build failure appears to be [this bug](https://github.com/iains/gcc-12-branch/issues/6), hopefully it will sort itself out soon.

PR #36, opened 2023-02-06
Add stride adaptor

tcbrindle:


tcbrindle:
Boo, CodeCov claims we're not testing the `detail::advance(seq, cur, 0)` cases, even though I'm pretty sure we are. Oh well.

PR #41, opened 2023-02-24
Try using GH Actions to make the single header

tcbrindle:


tcbrindle:
This isn't actually going to run 'til I merge it, so....

PR #45, opened 2023-04-12
Remove projections (!)

tcbrindle:
I like projections in std::ranges. But they come don't come for free: there is a (small?) overhead in terms of compile times, and an added burden in terms of specification, concept formulation and documentation.

What this PR does is add two new function adaptors/combinators which provide the equivalent functionality to projection arguments to algorithms, but in an opt-in basis. So instead of saying, for example

```cpp
flux::sort(vec, std::less{}, &T::member);
```

you now say

```cpp
flux::sort(vec, flux::proj(std::less{}, &T::member));
```

This arguably makes it more clear what's actually happening -- that we're modifying the comparator function object.

The `proj` adaptor takes an n-ary callable `F` and a unary projection `P` and returns a new n-ary callable which, when invoked with arguments `args...`, calls `F(P(args)...)` -- that is, it applies the projection to each argument before calling `F`.

This covers most uses of projections in the library. The exceptions are algorithms which take two sequences and two (different) projections -- `equal()` and `compare()` for example. For this case we have another adaptor `proj2` (which should really be called `binary_project` or something like that, but it's a bit too long). This takes a binary callable `F` and two unary projections `L` and `R`, and constructs a binary callable which, when invokes with arguments `a` and `b` calls `F(L(a), R(b))` -- that is, it applies the left and right projection functions to the left and right arguments. (Theoretically we could make it all variadic and have N projections for an n-ary callable, but we only ever need two in Flux.)

Both `proj` and `proj2` are actually aggregate classes used via CTAD, meaning you can use designated initialisers to say, for example

```cpp
flux::equal(seq1, seq2, flux::proj2{.fn = std::equal_to{}, .lhs = &T::member, .rhs = &U::member});
```

which I actually think reads quite well.

### DIFF
--- a/include/flux/core/concepts.hpp
+++ b/include/flux/core/concepts.hpp
@@ -301,15 +301,6 @@ concept derived_from_inline_sequence_base = requires(T t) {
 
 } // namespace detail
 
-/*
- * More useful concepts and associated types
- */
-template <typename Proj, sequence Seq>
-using projected_t = std::invoke_result_t<Proj&, element_t<Seq>>;
-
-template <typename Pred, typename Seq, typename Proj>
-concept predicate_for =
-    std::predicate<Pred&, std::invoke_result_t<Proj&, element_t<Seq>>>;
 
 /*
  * Default sequence_traits implementation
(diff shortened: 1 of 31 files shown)


tcbrindle:
CodeCov isn't happy, but I think that's because we've removed some lines that were hit in pqdsort.hpp, so the few that we miss now count for a greater percentage of the total

PR #48, opened 2023-04-21
Add example for simple config parser

jnytra:


tcbrindle:
All checks passed! ✅

This looks really good. It's great to see a more "real-world" example of Flux in action, thank you!

example/config_parser.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+token_t parse_line(const std::string& line)
+{
+    if (line.starts_with("#"))
+        return comment_t{line.substr(1)};
+    else if (line.starts_with("[") and line.ends_with("]"))
+        return section_t{line.substr(1, line.size() - 2)};
+
+    auto items = flux::from(line).split_string("=").to<std::vector>();
```
tcbrindle, line 62: It's a bit of a shame that we need to convert to a vector here rather than being able to use the sequence directly, but it does seem to be the tidiest way to do it and it also shows off the `to()` function so it's fine

tcbrindle (approved):

example/config_parser.cpp
jnytra: I agree it's not the most efficient way, but for demonstration purposes I'd say sufficient.

jnytra:
> All checks passed! ✅
> 
> This looks really good. It's great to see a more "real-world" example of Flux in action, thank you!
> 
> (Also, I'm very impressed with how you're using the library 100% correctly despite there not being much documentation at the moment, well done for working it out)

Thanks for feedback. Code is very intuitive and unit tests are the best documentation. :-)

example/config_parser.cpp
tcbrindle: Yep, definitely

PR #50, opened 2023-04-26
Add calendar example

jnytra:


tcbrindle:
This is amazing! I wanted to write a calendar example at some point but I thought I'd need an `interleave` adaptor like the Range-V3 version uses.

I should have time to do a full review over the weekend (and work out why the Windows builds are failing, if you don't figure it out first), but at first glance this looks really good.

example/calendar.cpp
```diff
@@ -0,0 +1,166 @@
...
+    auto month = std::string{months[static_cast<unsigned>(ymd{d}.month()) - 1]};
+    auto title = month + '-' + year;
+    auto pad_size = (week_str_size - title.size()) / 2;
+    auto aligned_title = (std::string(pad_size, ' ') + title);
+    aligned_title.resize(week_str_size, ' ');
+    return aligned_title;
+};
+
+auto week_to_string = [](auto&& w) { 
+    auto week_str = flux::from(FLUX_FWD(w))
```
tcbrindle, line 65: I don't think this needs the extra `flux::from` here -- we should just be able to say 

```cpp
FLUX_FWD(w)
    .map(day_to_string)
    .flatten()
    .template to<std::string>();
```

example/calendar.cpp
```diff
@@ -0,0 +1,166 @@
...
+                     .map(day_to_string)
+                     .flatten()
+                     .template to<std::string>();                     
+    bool first_week_in_month = ymd(*flux::front(w)).day() == 1d;
```
tcbrindle, line 69: This uses `w` after it's been forwarded/moved in the statement above -- this line needs to go first

example/calendar.cpp
```diff
@@ -0,0 +1,166 @@
...
+auto to_week_lines = [](auto&& month) { 
+    auto month_str = month_name(*flux::front(month));
+    auto week_lines = flux::from(FLUX_FWD(month))
+        .chunk_by(week_num)
+        .map(week_to_string);
+    std::vector<std::string> pad_lines(max_weeks_in_month - flux::count(week_lines), std::string(week_str_size, ' '));
+    return flux::chain(flux::single(std::move(month_str)), 
+                       flux::from(std::move(week_lines)), 
+                       flux::from(std::move(pad_lines)),
+                       flux::single(std::string(week_str_size, row_sep)));
```
tcbrindle, line 87: These don't need the extra `flux::from`

```suggestion
    return flux::chain(flux::single(std::move(month_str)), 
                       std::move(week_lines), 
                       std::move(pad_lines),
                       flux::single(std::string(week_str_size, row_sep)));
```

example/calendar.cpp
```diff
@@ -0,0 +1,166 @@
...
+    {
+        std::string padding(week_str_size - week_str.size(), ' ');
+        week_str = first_week_in_month ? padding + week_str : week_str + padding;
+    }
+    return week_str;
+};
+
+auto to_week_lines = [](auto&& month) { 
+    auto month_str = month_name(*flux::front(month));
+    auto week_lines = flux::from(FLUX_FWD(month))
```
tcbrindle, line 80: Again, I don't think this needs the `flux::from`

example/calendar.cpp
```diff
@@ -0,0 +1,166 @@
...
+    std::vector<std::string> pad_lines(max_weeks_in_month - flux::count(week_lines), std::string(week_str_size, ' '));
+    return flux::chain(flux::single(std::move(month_str)), 
+                       flux::from(std::move(week_lines)), 
+                       flux::from(std::move(pad_lines)),
+                       flux::single(std::string(week_str_size, row_sep)));
+};
+
+auto to_columns = [](auto&& month_chunk)
+{
+    auto n_rows = flux::count(*month_chunk.front());
```
tcbrindle, line 92: I think you should be able to say `month_chunk.front()->count()` which possibly looks a bit nicer

example/calendar.cpp
```diff
@@ -0,0 +1,166 @@
...
+    return flux::from(FLUX_FWD(month_chunk)).fold([](auto s, auto&& n) 
+    {
+        return flux::zip(flux::from(s), flux::from(FLUX_FWD(n)).template to<std::vector>())
+                    .map([](const auto& v) {
+                        return std::get<0>(v) + std::get<1>(v) + col_sep;
+                    }).template to<std::vector>();
+    }, std::vector<std::string>(n_rows, col_sep));
```
tcbrindle, line 99: This looks very clever, but to be honest, I find it quite difficult to follow what's going on here with a zip inside a fold.  Can I suggest moving the lambda out into a separate function and giving the variables meaningful names (rather than just `s` and `n`) to make it a bit easier to read?

tcbrindle (changes requested):
This is very impressive work!

My only worry is that it's a bit hard to understand what's going on as an example of how to use Flux, particularly with all the uses of `auto` parameters and deduced return types on the various functions. Could you add some comments (similar the ones the Range-V3 calendar example uses) to give a hint as to what types the lambdas are dealing with? For example

```cpp
/*
  In: sequence<sequence<sys_day>>
  Out: std::string
*/
auto week_to_string = [](auto&& w) -> std::string {
    ...
};
```

jnytra:
Thanks so much for review. I will try to fix your comments as soon as possible.

example/calendar.cpp
```diff
@@ -0,0 +1,165 @@
...
+    std::vector<std::string> pad_lines(max_weeks_in_month - flux::count(week_lines), std::string(week_str_size, ' '));
+    return flux::chain(flux::single(std::move(month_str)), 
+                       std::move(week_lines), 
+                       std::move(pad_lines),
+                       flux::single(std::string(week_str_size, row_sep)));
+};
+
+auto append_column = [](std::vector<std::string>&& months_per_line, flux::sequence auto&& month) {
+    return flux::zip(flux::from(std::move(months_per_line)), FLUX_FWD(month).template to<std::vector>())
+        .map([](const auto& v) { return std::get<0>(v) + std::get<1>(v) + col_sep; })
```
tcbrindle, line 95: `zip` with two arguments returns a prvalue pair of references, so you could say 

```suggestion
        .map([](auto p) { return p.first + p.second + col_sep; })
```

which looks a bit nicer than `std::get<N>`

example/calendar.cpp
```diff
@@ -0,0 +1,165 @@
...
+                        .map(week_to_string);
+    std::vector<std::string> pad_lines(max_weeks_in_month - flux::count(week_lines), std::string(week_str_size, ' '));
+    return flux::chain(flux::single(std::move(month_str)), 
+                       std::move(week_lines), 
+                       std::move(pad_lines),
+                       flux::single(std::string(week_str_size, row_sep)));
+};
+
+auto append_column = [](std::vector<std::string>&& months_per_line, flux::sequence auto&& month) {
+    return flux::zip(flux::from(std::move(months_per_line)), FLUX_FWD(month).template to<std::vector>())
```
tcbrindle, line 94: We don't need the `flux::from` around the first argument here

```suggestion
    return flux::zip(std::move(months_per_line), FLUX_FWD(month).template to<std::vector>())
```

Also, it makes me sad that we need to convert `month` to a vector here for this to work... I need to understand why this breaks if you just pass `month`,  because I think it should work okay

example/calendar.cpp
```diff
@@ -0,0 +1,165 @@
...
+}
+
+app_args_t parse_args(int argc, char** argv) {
+    app_args_t result;
+    std::string_view app_name = argv[0];
+    std::vector<std::string> args(std::next(argv), std::next(argv, argc));
+
+    using parse_func_t = std::function<void(std::string)>;
+    std::map<std::string, parse_func_t> args_map {
+        {"--help"s, [&](std::string val) { print_help_and_exit(app_name); }},
```
tcbrindle, line 127: I get a very silly "unused argument" warning here

```suggestion
        {"--help"s, [&](std::string) { print_help_and_exit(app_name); }},
```

example/calendar.cpp
```diff
@@ -0,0 +1,165 @@
...
+// Out: std::vector<std::string> (months organized in columns)
+auto to_columns = [](flux::sequence auto&& month_chunk) {
+    auto n_rows = month_chunk.front()->count();
+    return FLUX_FWD(month_chunk).fold(append_column, std::vector<std::string>(n_rows, col_sep));
+};
+
+const auto current_year = ymd{floor<days>(system_clock::now())}.year();
+
+struct app_args_t {
+    bool help;
```
tcbrindle, line 109: Is the `help` member used anywhere?

tcbrindle (changes requested):
Thanks for making the changes, it's much easier to follow with the comments :) 

This looks in really good shape, there are just a couple of little things I noticed but then we should be good to merge it.

example/calendar.cpp
jnytra: Good point. Fixed!

example/calendar.cpp
jnytra: No, thanks.

example/calendar.cpp
jnytra: I changed to `{"--help"s, [&]([[maybe_unused]] std::string) { print_help_and_exit(app_name); }},`. It is ok for you?

example/calendar.cpp
jnytra: I removed unneeded `flux::from`.

example/calendar.cpp
tcbrindle: I think just removing the name `val` is probably enough to keep the compiler happy, but `[[maybe_unused]]` doesn't do any harm

tcbrindle (approved):

jnytra:
There is one final problem with the MacOS pipeline. The libc++ standard library doesn't implement an increment operator for `sys_days` 😞. We can do it with `ifdef` like this: https://godbolt.org/z/5P8oM4MK5. What do you think?

tcbrindle:
> There is one final problem with the MacOS pipeline. The libc++ standard library doesn't implement an increment operator for `sys_days` 😞. We can do it with `ifdef` like this: https://godbolt.org/z/5P8oM4MK5. What do you think?

It's a shame we need to do this, but I think it's the best solution.

We should probably add a version check too, because hopefully libc++ 17 will add these operators so it won't need the workaround. Something like the following I think would work:

```cpp
// Workaround: libc++16 does not support C++20 chrono::time_point::operator++
#ifdef _LIBCPP_VERSION
#if _LIBCPP_VERSION < 170000
namespace std::chrono {
    sys_days& operator++(sys_days& d) {
        return d += days{1};
    }

    sys_days operator++(sys_days& d, int) {
        return std::exchange(d, d + days{1});
    }
}
#endif
#endif
```

jnytra:
Okay, I simplified it like this:
```c++
#if defined _LIBCPP_VERSION and _LIBCPP_VERSION < 170000
```

tcbrindle:
> Okay, I simplified it like this:
> 
> ```c++
> #if defined _LIBCPP_VERSION and _LIBCPP_VERSION < 170000
> ```

Heh, last time I tried it MSVC didn't like having two preprocessor conditions on one line which is why I suggested the two nested `#if`s, but it looks like they've fixed it 👍

tcbrindle:
This looks great, thank you very much!

tcbrindle (approved):

PR #51, opened 2023-04-26
Add merge intervals example

jnytra:


example/merge_intervals.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    return s << '(' << i.begin << ',' << i.end << ')';
+}
+
+bool is_overlapped(interval_t a, interval_t b)
+{
+    return a.end >= b.begin;
+}
+
+auto merge = [](auto seq) -> interval_t
```
tcbrindle, line 27: You could add a constraint to the lambda auto param here, i.e. `[](flux::sequence auto) -> interval_t` to make it clear what kind of parameter the lambda is expecting

example/merge_intervals.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    auto begin = flux::front(seq)->begin;
+    auto end = flux::from(seq).max(flux::proj(std::less<>{}, &interval_t::end))->end;
```
tcbrindle, line 30: It's slightly less typing to use the free-function form of `max` here, i.e.

    auto end = flux::max(seq, flux::proj(std::less<>{}, &interval_t::end))->end;

example/merge_intervals.cpp
```diff
@@ -0,0 +1,51 @@
...
+    std::vector<interval_t> intervals = {{2, 4}, {7, 9}, {11, 13}, {6, 7}, {0, 3}};
+
+    // sort intervals according begin
+    flux::sort(intervals, flux::proj(std::less<>{}, &interval_t::begin));
+
+    auto merged = flux::from(intervals)
+                       .chunk_by(is_overlapped)
+                       .map(merge);
+                
+    flux::write_to(merged, std::cout);
```
tcbrindle, line 45: It doesn't make a whole lot of difference, but you should be able to add the `write_to` to the end of the pipeline here to avoid the intermediate variable:

    flux::from(intervals)
        .chunk_by(is_overlapped)
        .map(merge)
        .write_to(std::cout);

should work I think?

example/merge_intervals.cpp
```diff
@@ -0,0 +1,51 @@
...
+    std::cout << std::endl;
+
+    // prints [(0,4), (6,9), (11,13)]
+
+    return 0;
```
tcbrindle, line 50: No need for `return 0` at the end of main, I usually leave it off :)

tcbrindle (changes requested):
A really nice little example, in particular how to use the new proj functions, thanks you!

There are a couple of little things that I've noted but they're not a big deal.

example/merge_intervals.cpp
jnytra: Good point.

example/merge_intervals.cpp
jnytra: It works. :-)

example/merge_intervals.cpp
```diff
@@ -0,0 +1,49 @@
...
+    auto end = flux::max(seq, flux::proj(std::less<>{}, &interval_t::end))->end;
+    return {begin, end};
+};
+
+int main()
+{
+    std::vector<interval_t> intervals = {{2, 4}, {7, 9}, {11, 13}, {6, 7}, {0, 3}};
+
+    // sort intervals according begin
+    flux::sort(intervals, flux::proj(std::less<>{}, &interval_t::begin));
```
tcbrindle, line 39: Do I remember rightly that C++20 allows you to omit the empty angle brackets after `less`? (I always forget this as I'm so used to writing them...)

```suggestion
    flux::sort(intervals, flux::proj(std::less{}, &interval_t::begin));
```

example/merge_intervals.cpp
```diff
@@ -0,0 +1,49 @@
...
+    auto begin = flux::front(seq)->begin;
+    auto end = flux::max(seq, flux::proj(std::less<>{}, &interval_t::end))->end;
+    return {begin, end};
+};
+
+int main()
+{
+    std::vector<interval_t> intervals = {{2, 4}, {7, 9}, {11, 13}, {6, 7}, {0, 3}};
+
+    // sort intervals according begin
```
tcbrindle, line 38: Typo in comment

```suggestion
    // sort intervals according to begin
```

tcbrindle (commented):
A couple of tiny things, sorry I didn't notice them before

tcbrindle (approved):

PR #55, opened 2023-04-28
Add documentation for drop adaptor

jnytra:


tcbrindle:
Thanks for this!

It occurs to me that we should probably decide, and then document what happens if the user passes a negative value to `drop()`, and then I need to update the actual implementation to match the documented behaviour! (This also applies to `take()`).

We could either raise a `runtime_error` if the user passes a negative value, or else clamp the value to zero (so we don't actually drop anything). Currently I think the former is probably the way to go.

docs/reference/adaptors.rst
```diff
@@ -104,6 +104,27 @@ Adaptors
 ..  function::
     auto drop(sequence auto seq, std::integral auto count) -> sequence auto;
 
+    Given a sequence :var:`seq` and a non-negative integral value :var:`count`, returns a new sequence which skips the first :var:`count` elements of :var:`seq`.
+
+    The returned sequence is always a :concept:`sequence`, with the same category of :var:`seq`. If :var:`seq` has fewer than :var:`count` elements, the adapted sequence is out of range.
```
tcbrindle, line 109: > The returned sequence is always a :concept:`sequence`, with the same category of :var:`seq`

I think that elsewhere I say "the returned sequence has the same capabilities as :var:`seq`".

> If :var:`seq` has fewer than :var:`count` elements, the adapted sequence is out of range.

"Out of range" is not quite right: it would be better to say "..., the returned sequence is empty."

example/docs/drop.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include <flux.hpp>
+
+#include <cassert>
+#include <vector>
+
+int main()
+{
+    std::vector vec{1, 2, 3, 4, 5};
+
+    auto dropped = flux::from(vec).drop(3);
```
tcbrindle, line 16: At the moment I'm only using the free-function versions of algorithms in examples -- I still need to decide the best way of documenting the member-function versions.

```suggestion
    std::vector vec{1, 2, 3, 4, 5};

    auto dropped = flux::drop(std::move(vec), 3);
```

example/docs/drop.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+    assert(flux::count(dropped) == 2);
```
tcbrindle, line 18: Could use `size()` here rather than `count()` to show that the adapted sequence is `sized` when the input sequence is

```suggestion
    assert(flux::size(dropped) == 2);
```

tcbrindle (changes requested):

docs/reference/adaptors.rst
jnytra: Thanks for corrections. Originally, I want to write "the returned sequence is empty", but this example not working:
https://godbolt.org/z/z88srjvx5

jnytra:
Raise a `runtime_error` or `invalid_argument` if the user passes a negative value, sound good.

docs/reference/adaptors.rst
tcbrindle: Oh dear, that's really bad :(

Drop and take were two of the first adaptors I wrote for Flux, it looks like I didn't do a very good job!

tcbrindle:
Let me fix the `drop` and `take` adaptors so they match what we'd like them to do, then I can merge the documentation

tcbrindle (approved):

PR #59, opened 2023-05-02
CI: test with GCC13

tcbrindle:
Now that GCC13 has been released, we should add it to our CI set up.


tcbrindle:
Reported code coverage seems to have gone down in many places (including a rather alarming 100% drop `take_while_adaptor`!). If I had to guess I'd say this is because GCC13 is now inlining more stuff at `-O0`, which is actually a good thing but means we'll have to put up with CodeCov complaining for a while

PR #61, opened 2023-05-02
Add word count example

jnytra:


example/word_count.cpp
```diff
@@ -0,0 +1,47 @@
...
+    return flux::fold(FLUX_FWD(seq), [](stats_t&& stats, auto val) 
+    {
+        stats.chars++;
+        if (not stats.is_last_space and std::isspace(val))
+            stats.words++;
+        if (val == '\n')
+            stats.lines++;
+        stats.is_last_space = std::isspace(val);
+        return FLUX_FWD(stats);
+    }, stats_t{});
```
tcbrindle, line 31: As the stats struct is quite large and needs to be copied for each element, I wonder if it would be better to use `for_each` here? i.e. something like

```cpp
    stats_t stats{};
    flux::for_each(FLUX_FWD(seq), [&](char val) {
        stats.chars++;
        if (not stats.is_last_space and std::isspace(val))
            stats.words++;
        if (val == '\n')
            stats.lines++;
        stats.is_last_space = std::isspace(val);
    });

    return stats;
```

(Or maybe it's no different?)

example/word_count.cpp
```diff
@@ -0,0 +1,47 @@
...
+    return s << stats.lines << ' ' << stats.words << ' ' << stats.chars;
+}
+
+int main()
+{
+    // Print newline, word, and byte counts from std input (like gnu wc)
+    std::noskipws(std::cin);
+    auto result = flux::from_istream<char>(std::cin)._(collect_stats);
+    if (not result.is_last_space)
+        result.words++;
```
tcbrindle, line 45: Could these lines be moved into the `collect_stats` function to make `main()` look even cleaner?

tcbrindle:
Another great little example, thank you!

tcbrindle:
I don't know why the CI checks aren't running though...

tcbrindle (approved):
Looks great, thanks!

PR #66, opened 2023-05-04
Add custom adaptor tutorial example

tcbrindle:


tcbrindle:
@jnytra If you can, would you mind looking through this tutorial/example and letting me know if anything is unclear or doesn't make sense?

jnytra:
This is absolutely amazing. This is exactly what I needed to write a new adaptor. I don't have any comments yet, but I plan to write a new adaptor according to this tutorial, so I'll see.

example/tutorials/custom_adaptor.cpp
```diff
@@ -0,0 +1,311 @@
...
+     * To implement the Flux sequence API, we use a public nested class named
+     * `flux_sequence_traits`.
+     *
+     * Alternatively we could provide a class template specialisation
+     *     flux::sequence_traits<repeat_elements_adaptor<Base>>
+     * at global namespace scope, but a nested class is usually much more
+     * convenient.
+     */
+
+    struct flux_sequence_traits {
```
nwad123, line 133: Change to `struct flux_sequence_traits : flux::default_sequence_traits {`

nwad123:
This tutorial is great, it's exactly what I needed to get started. One issue I ran into is that `repeat_elements_adaptor::flux_sequence_traits` doesn't inherit from `flux::default_sequence_traits` which prevented the example from building.  Changing line 133 from:

```cpp
struct flux_sequence_traits {
```

to 

``` cpp
struct flux_sequence_traits : flux::default_sequence_traits {
```

resolves the issue. I can't add a commit to this pr, but I didn't think that this change warranted a completely new pull request. Thanks again for the tutorial!

PR #67, opened 2023-05-05
Add moving average example

jnytra:


example/moving_average.cpp
```diff
@@ -0,0 +1,68 @@
...
+    return std::move(win);
+};
+
+int main() {
+    std::vector intervals = {1, 5, 6, 1, 2, 9, 7, -1, 0};
+
+    // compute moving average by scan adaptor (more effective for large windows)
+    auto ma = flux::from(intervals)
+        .scan(sliding_window, sliding_window_t(3))
+        .map(flux::proj(&sliding_window_t::average))
```
tcbrindle, line 48: I think you should be able to pass the mem fn ptr directly to `map`, without needing to use `proj`:

```suggestion
        .map(&sliding_window_t::average)
```

tcbrindle:
An example using `scan` already? I only merged it yesterday! :D

This is another great little example, thanks once again

jnytra:
I accidentally created a pull request from my main. Sorry for that. I am closing this one and creating a new one with fixes.

PR #68, opened 2023-05-05
Add moving average example

jnytra:


example/moving_average.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include <cstddef>
+#include <cassert>
+#include <deque>
+#include <vector>
+
+struct sliding_window_t
+{
+    explicit sliding_window_t(std::size_t N): size(N) {
+    }
```
tcbrindle, line 15: This is very nit-picky, but uppercase `N` here makes it look like a template parameter: I normally use lowercase names for runtime variables. Also for a constructor that can fit on a single line, I'd usually format it like this:

```suggestion
    explicit sliding_window_t(std::size_t n) : size(n) {}
```

jnytra: Thanks, fixed.

tcbrindle (approved):

PR #70, opened 2023-05-09
Make sure we actually run tests at runtime

tcbrindle:
Because almost the entire library is constexpr, the majority of our tests look like this:

```cpp
constexpr bool test_xxx() {
    // test code
    return true;
}
static_assert(test_xxx());

TEST_CASE("xxx") {
    REQUIRE(test_xxx());
}
```

That is, we do a constexpr test during compilation (which has the added benefit of testing for many guises of UB), and then again at runtime using Catch.

Unfortunately, compilers are getting increasingly smart at noticing that they've already calculated the result of `test_xxx()`, so they can optimise the run-time version to just `REQUIRE(true)`, which isn't exactly in the spirit of what we want as it means we never actually execute the tests at runtime! It also means that the tested functions never show up in our code coverage measurements.

As a workaround, this commit adds an otherwise-useless `template <bool = true>` to every `test_xxx()` function. Then in the runtime version we explicitly call `test_xxx<false>()` to force a separate instantation which should hopefully show up in CodeCov.

This probably won't do wonders for our compile times, but we'll see...

### DIFF
--- a/test/test_adjacent.cpp
+++ b/test/test_adjacent.cpp
@@ -25,6 +25,7 @@ constexpr bool tuple_equal(Lhs const& lhs, Rhs const& rhs)
             impl(std::make_index_sequence<std::tuple_size_v<Lhs>>{});
 }
 
+template <bool = true>
 constexpr bool test_pairwise()
 {
     // Basic pairwise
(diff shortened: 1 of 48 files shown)


tcbrindle:
Well, that was a complete waste of time.

It looks like GCC is perfectly capable of executing `test_xxx<false>()` at compile-time and using that, even though we didn't explicitly ask for it.

Ah well, never mind.

Issue #80, opened 2023-05-26
CodeCov action is broken


tcbrindle:
It looks like Homebrew has updated the version of `lcov` from 1.16 to 2.0, which has broken things.

PR #81, opened 2023-05-26
Fix CodeCov action

tcbrindle:
Homebrew has updated `lcov` from v1.16 to 2.0, and it looks like we're now getting lots of warnings and errors during data collection.


tcbrindle:
Well, that was less painful than expected. Looks like this fixes #80

PR #85, opened 2023-06-21
Add mask adaptor

### LINKED ISSUE #84: Add compress adaptor
In this issue, I suggest adding `compress` sequence adaptor. 

Compress adaptor filters elements from the input sequence, returning only those that have a corresponding element in selectors that evaluates to true. Stops when either the input sequence or selectors sequences has been exhausted. Roughly equivalent to:

```cpp
flux::generator compress(seq, selectors) {
    // compress("ABCDEF", {1,0,1,0,1,1}) --> A C E F
    for (auto [v, s]: flux::zip(seq, selectors)) {
        if (s) {
            co_yield v;
        }
    }
}
```

tcbrindle:
Given a sequence of values and a sequence of selectors, `select_by(vals, selectors)` yields those elements of `vals` for which the corresponding element of `selectors` is `true` (after conversion to `bool`).

Python itertools calls this "compress", but I'm not overly keen on that name. On the other hand, `select_by` isn't ideal either -- `chunk_by` takes a predicate, but this takes a second sequence, so there's an inconsistency there. I did consider just calling it `select()`, but then [selection algorithms](https://en.wikipedia.org/wiki/Selection_algorithm) are a thing...

Fixes #84


jnytra:
The code looks nice. As for the name of the adapter, I just prefer `select()`, but here are some other names that came to mind:

* **mask**: This name implies that the function applies a masking operation on the input iterator, retaining only the elements that meet the specified condition.

* **subset**: This name suggests that the function creates a subset of the original iterator by selectively including or excluding elements based on a given criterion.

* **filter_true**: This name reflects the behavior of the function, where it filters elements based on a truth value. It conveys the idea that it selects elements that evaluate to `true`.

tcbrindle:
Thanks for the name suggestions. I agree `select` is the nicest, but I quite like `mask` as well, and that doesn't have an existing meaning

tcbrindle:
I had a chat with a few folks at C++ on Sea about the name, and I was reminded that Posix `select()` is a thing, which might be confusing. Sean Parent suggested the name `mask()`, so that seems like a good one to go for.

tcbrindle:
The good news is that the Windows CI problems seem to have resolved themselves. The bad news is that now the MacOS CI isn't working because of problems with Homebrew *sigh*. Hopefully this will get sorted out too and we can merge this.

jnytra:
I also like the `mask()` version the most.

PR #92, opened 2023-07-11
Implement internal iteration for `cartesian_product`

brycelelbach:
This PR implements internal iteration for `cartesian_product[_with]`, which enables compiler vectorization of multidimensional iteration patterns constructed with `cartesian_product[_with]` (such as [this one](https://godbolt.org/z/7oshP5dTb)).

Core tasks:

* [x] Add internal iteration to `cartesian_product`.
  * [x] Correct `cartesian_product` internal iteration to invoke the predicate with a tuple of elements, not a tuple of indices (review feedback).
* [ ] Add internal iteration to `cartesian_product_with`.
* [x] Fix `cartesian_product[_with]` random access indexing.
  * [x] Don't treat cursors as arbitrary numeric types (review feedback).
* [x] Improve testing for `cartesian_product[_with]` in general.
  * [x] Test `cartesian_product` with 3 underlying sequences.
  * [x] Test `cartesian_product` with less than 2 and more than 3 underlying sequences.
  * [x] Test `cartesian_product` random access indexing.
  * [x] Test `cartesian_product` `for_each_while` short circuiting (review feedback).
  * [x] Test `cartesian_product[_with]` with `for_each[_while]`.
  * [x] Correct element type tests for `cartesian_product` of `ints` to fix MSVC build (review feedback).
* [x] Add benchmarks and examples for multi-dimensional iteration with `cartesian_product` and `iota`.
* [ ] Add descriptive alias for `cartesian_product` of `iota` multi-dimensional iteration pattern.
* [x] Miscellaneous drive-by fixes and improvements.
  * [x] Change `STATIC_CHECK` to throw instead of returning to improve compile-time diagnostics.
  * [x] Disable `word_count` example CTest as it will always hang waiting for stdin.


tcbrindle:
Thanks very much for the drive-by fixes, they look great!

In `for_each_while_impl`, we need to eventually call the predicate with a tuple of *elements*, not a tuple of *cursors*. It looks like you are building up a tuple of elements (despite the param name), but constructing the tuple on line 271 doesn't look right to me -- I think it should be `element_t` rather than `cursor_t`?

(This happens to work for the particular case of a cartesian product of iotas, because for an iota sequence the element type and the cursor type happen to be the same.) 

Other than that I think this looks like a reasonable approach -- I wish the `keep_going` variable wasn't necessary at every level, but I can't think of a way to get rid of it, and if it doesn't affect vectorisation then I guess we're okay.

brycelelbach:
> In for_each_while_impl, we need to eventually call the predicate with a tuple of elements, not a tuple of cursors. It looks like you are building up a tuple of elements (despite the param name), but constructing the tuple on line 271 doesn't look right to me -- I think it should be element_t rather than cursor_t?
> 
> (This happens to work for the particular case of a cartesian product of iotas, because for an iota sequence the element type and the cursor type happen to be the same.)

Ah-ha, I think I suffered from a misunderstanding here. I believe you are correct.

I'm surprised the tests didn't uncover this, but I just realized we don't have any tests that actually call `for_each` on `cartesian_product`. I was just testing with my benchmark.

test/test_cartesian_product.cpp
```diff
@@ -110,6 +111,167 @@ constexpr bool test_cartesian_product()
...
+
+        static_assert(flux::sequence<C const>);
+        static_assert(flux::multipass_sequence<C const>);
+        static_assert(flux::bidirectional_sequence<C const>);
+        static_assert(flux::random_access_sequence<C const>);
+        static_assert(not flux::contiguous_sequence<C const>);
+        static_assert(flux::bounded_sequence<C const>);
+        static_assert(flux::sized_sequence<C const>);
+
+        static_assert(std::same_as<flux::element_t<C>, std::tuple<long, long, long>>);
```
tcbrindle, line 207: `C` is `cartesian_product(ints(), ints(), ints())`, so its element type is `std::tuple<distance_t, distance_t, distance_t>`, where `distance_t` is our alias for `ptrdiff_t`. On 32-bit Windows `ptrdiff_t` is just `int`, hence the Windows static assert failures. (But it would fail on 32-bit Linux too, if tested that.)

One possible solution:

```suggestion
        static_assert(std::same_as<flux::element_t<C>, std::tuple<flux::distance_t, flux::distance_t, flux::distance_t>>);
```

(And the same for the other static asserts)

brycelelbach:
> Other than that I think this looks like a reasonable approach -- I wish the keep_going variable wasn't necessary at every level, but I can't think of a way to get rid of it, and if it doesn't affect vectorisation then I guess we're okay.

I was a little bit concerned that the control flow of this would cause issues, but it seems to be fine.

The one idea I had for working around this was to not use `flux::for_each_while` to iterate each of the underlying sequences, however that seems like a gross violation of the basic principles here. If we did that, it would break more deeply nested internal iteration (for example, `cartesian_product(filter(...), ...)`).

brycelelbach:
> I'm surprised the tests didn't uncover this, but I just realized we don't have any tests that actually call `for_each` on `cartesian_product`. I was just testing with my benchmark.

Ironically, the `unpack` test that you added does a `for_each` on `cartesian_product`, but it's on `ints` so it doesn't catch this, and I didn't run it before committing anyways.

test/test_cartesian_product.cpp
brycelelbach: Fixed in 62a25c31de697cff4fa163e48eda1409a3c81059

example/word_count.cpp
tcbrindle, line 1: I guess this was unintentional whitespace changes from your editor?

brycelelbach:
> Ah-ha, I think I suffered from a misunderstanding here. I believe you are correct.
> 
> I'm surprised the tests didn't uncover this, but I just realized we don't have any tests that actually call `for_each` on `cartesian_product`. I was just testing with my benchmark.

I'm having trouble writing a test that covers this issue. Any ideas?

include/flux/op/cartesian_product.hpp
```diff
@@ -88,21 +89,36 @@ struct cartesian_product_traits_base {
...
+            return cur;
+
         auto& base = std::get<I>(self.bases_);
+        auto& this_index = std::get<I>(cur);
+        auto this_size = flux::size(base);
 
-        auto this_sz = flux::size(base);
-        auto this_offset = offset  % this_sz;
-        auto next_offset = offset/this_sz;
+        this_index += offset;
```
tcbrindle, line 101: Use `checked_add` here to avoid potential UB on signed overflow

```suggestion
        this_index = num::checked_add(this_index, offset);
```

(Yeah, it's ugly, I know :( )

include/flux/op/cartesian_product.hpp
```diff
@@ -88,21 +89,36 @@ struct cartesian_product_traits_base {
...
     template <std::size_t I, typename Self>
     static constexpr auto ra_inc_impl(Self& self, cursor_type<Self>& cur, distance_t offset)
-    -> cursor_type<Self>&
+        -> cursor_type<Self>&
     {
+        if (offset == 0)
+            return cur;
+
         auto& base = std::get<I>(self.bases_);
+        auto& this_index = std::get<I>(cur);
```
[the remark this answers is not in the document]
brycelelbach: Ah, yah, I was a bit worried about this.

example/word_count.cpp
brycelelbach: Ah, I think it was stripping trailing whitespace from the end of the line. Want me to remove that change?

include/flux/op/cartesian_product.hpp
```diff
@@ -216,6 +255,37 @@ struct sequence_traits<detail::cartesian_product_adaptor<Bases...>>
...
+                                              PartialCursor&&... partial_cursor)
+        -> std::tuple<bool, detail::cartesian_product_partial_cursor_t<Self, I>>
+    {
+        // We need to iterate right to left.
+        if constexpr (I == sizeof...(Bases) - 1) {
+            bool keep_going = true;
+            auto this_current = flux::for_each_while(std::get<I>(self.bases_),
+                [&](auto&& elem) {
+                    keep_going = std::invoke(func,
+                        cursor_t<Self>(FLUX_FWD(partial_cursor)..., FLUX_FWD(elem)));
```
brevzin, line 271: This took me so long to figure out what was going on, because it seems obviously wrong but the tests still pass. But that's because the only cartesian product test for this uses `flux::ints` where this just happens to be the right thing.

The issues here are:
* `partial_cursor...` are not actual partial cursors, they're elements
* we need to pass an element (i.e. a tuple of elements) to `func`, not a cursor.

So if you try to do something like this, it doesn't compile:
```cpp
flux::cartesian_product(flux::ints(0, 3), flux::from_range(std::vector{10, 20, 30}))
    .for_each([](auto&&){});
```

Have to do this instead:

```suggestion
                        element_t<Self>(FLUX_FWD(partial_cursor)..., FLUX_FWD(elem)));
```

And then rename `partial_cursor` to `elems` or something.

brycelelbach:
> In `for_each_while_impl`, we need to eventually call the predicate with a tuple of _elements_, not a tuple of _cursors_. It looks like you are building up a tuple of elements (despite the param name), but constructing the tuple on line 271 doesn't look right to me -- I think it should be `element_t` rather than `cursor_t`?
> 
> (This happens to work for the particular case of a cartesian product of iotas, because for an iota sequence the element type and the cursor type happen to be the same.)

Tests added in a240dcc8d6aa4a1023e28ee94d83d3d68d640d82, fixed in 02e73b3c0d97a414f600dfe5391832cf12a67809.

tcbrindle:
> > Ah-ha, I think I suffered from a misunderstanding here. I believe you are correct.
> > I'm surprised the tests didn't uncover this, but I just realized we don't have any tests that actually call `for_each` on `cartesian_product`. I was just testing with my benchmark.
> 
> I'm having trouble writing a test that covers this issue. Any ideas?

I think what you want is some sequences where the element type and cursor type are distinct and can't be converted to one another? What about something like:

```cpp
auto seq1 = flux::ints(1).take(3);
auto seq2 = std::array<std::string_view, 2>{"aaa", "bbb"};

auto cart = flux::cartesian_product(seq1, seq2);

auto cur = flux::find_if(cart, flux::unpack([](auto i, auto s) { return i == 2 && s == "bbb"; }));

STATIC_CHECK(std::get<0>(cart[cur]) == 2);
STATIC_CHECK(std::get<1>(cart[cur]) == "bbb");
```

include/flux/op/cartesian_product.hpp
brycelelbach: Thanks - Tristan pointed this out early. I added tests for this in a240dcc8d6aa4a1023e28ee94d83d3d68d640d82, and fixed it similar to how you described in 02e73b3c0d97a414f600dfe5391832cf12a67809.

brevzin: There's no test for short-circuiting - if you just `return true;` in all the relevant places, all the tests pass.

example/word_count.cpp
brycelelbach: Fixed.

include/flux/op/cartesian_product.hpp
brycelelbach: Fixed in ba910d7c432ec5b70ba58ed428b1cbbb48360c25.

include/flux/op/cartesian_product.hpp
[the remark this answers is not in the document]
brycelelbach: Fixed in ba910d7c432ec5b70ba58ed428b1cbbb48360c25.

include/flux/op/cartesian_product.hpp
brycelelbach: Adding a test for short circuiting seems wise.

brycelelbach: Test added in 155a5ddaa8a6d0780d5b09ffe134bb694b990fc.

include/flux/op/cartesian_product.hpp
```diff
@@ -193,6 +209,29 @@ struct cartesian_product_traits_base {
...
+{
+    using type = decltype(std::tuple_cat(
+        std::declval<std::tuple<std::tuple_element_t<I::value, cursor_t<Self>>>>(),
+        std::declval<typename cartesian_product_partial_cursor<Self,
+            std::integral_constant<std::size_t, I::value + 1>>::type>()));
+};
+
+template <typename Self, std::size_t I>
+using cartesian_product_partial_cursor_t =
+    typename cartesian_product_partial_cursor<Self, std::integral_constant<std::size_t, I>>::type;
```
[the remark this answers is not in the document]
brycelelbach: That looks better, thanks. Change made in 56cdf32747940f95f5b1fcf3e018f67b8249d48d.

test/test_cartesian_product.cpp
```diff
@@ -17,9 +24,59 @@ namespace {
...
+        static_assert(flux::sized_sequence<C const>);
+
+        static_assert(std::same_as<flux::element_t<C>, std::tuple<int&>>);
+        static_assert(std::same_as<flux::value_t<C>, std::tuple<int>>);
+        static_assert(std::same_as<flux::rvalue_element_t<C>, std::tuple<int&&>>);
+
+        // FIXME: These currently fail - they're actually `tuple<const int>` etc. Is that correct?
+        //static_assert(std::same_as<flux::element_t<C const>, std::tuple<int&>>);
+        //static_assert(std::same_as<flux::value_t<C const>, std::tuple<int>>);
+        //static_assert(std::same_as<flux::rvalue_element_t<C const>, std::tuple<int&&>>);
```
brycelelbach, line 56: @tcbrindle I added a test for a 1D `cartesian_product` and was surprised that these tests failed. Thoughts?

benchmark/multidimensional_memset_benchmark_kernels.cpp
```diff
@@ -0,0 +1,24 @@
...
+            A[i + j * N] = 0.0;
+}
+
+void memset_2d_flux_cartesian_product_iota(double* A, std::size_t N, std::size_t M)
+{
+    flux::cartesian_product(flux::iota(0LU, N), flux::iota(0LU, M))
+        .for_each(flux::unpack([&] (auto i, auto j) {
+            A[i + j * N] = 0.0;
+        }));
+}
```
brycelelbach, line 24: This still doesn't work as intend - LLVM fails to vectorize it, leading this code to be 10x slower than it's C-style counterpart.

However, I'm pretty sure we can eventually crack this. The next thing I'm going to try is a separate customization path for `for_each`.

benchmark/multidimensional_memset_benchmark_kernels.cpp
```diff
@@ -0,0 +1,24 @@
...
+void memset_2d_reference(double* A, std::size_t N, std::size_t M)
+{
+    for (std::size_t j = 0; j != M; ++j)
+        for (std::size_t i = 0; i != N; ++i)
+            A[i + j * N] = 0.0;
+}
+
+void memset_2d_flux_cartesian_product_iota(double* A, std::size_t N, std::size_t M)
+{
+    flux::cartesian_product(flux::iota(0LU, N), flux::iota(0LU, M))
```
tcbrindle, line 20: The `iota()` calls here are tripping up 32-bit Windows because `size_t` isn't unsigned long there, so we get ambiguous deduction (`iota()` expects the same type for both arguments)

```suggestion
    flux::cartesian_product(flux::iota(std::size_t{0}, N), flux::iota(std::size_t{0}, M))
```

benchmark/multidimensional_memset_benchmark_kernels.cpp
tcbrindle, line 24: To match the reference version above, this should be

```suggestion
void memset_2d_flux_cartesian_product_iota(double* A, std::size_t N, std::size_t M)
{
    flux::cartesian_product(flux::iota(0LU, M), flux::iota(0LU, N))
        .for_each(flux::unpack([&] (auto j, auto i) {
            A[i + j * N] = 0.0;
        }));
}
```

which (for me) gives the same runtime for both

brycelelbach: Fixed slightly differently in 4132087e2b68de4ad69c0643e1cbaff2b67b5ecd - I switched the iteration order of the reference, instead of switching the iteration order of the flux variant.

include/flux/op/cartesian_product.hpp
```diff
@@ -193,6 +209,19 @@ struct cartesian_product_traits_base {
...
+template <typename T, std::size_t I, std::size_t... Js>
+struct drop_impl<T, I, std::index_sequence<Js...>> {
+    using type = std::tuple<std::tuple_element_t<I + Js, T>...>;
+};
+
+template <typename T, std::size_t I>
+using drop = typename drop_impl<T, I, std::make_index_sequence<std::tuple_size_v<T> - I>>::type;
+
+template <typename Self, std::size_t I>
+using cartesian_product_partial_cursor_t = drop<cursor_t<Self>, I>;
```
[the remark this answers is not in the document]
brycelelbach: Yah, `drop` isn't an optimal name here. I'll just call it `cartesian_product_tuple_drop` to avoid ambiguity.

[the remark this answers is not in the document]
brycelelbach: Fixed in 23df72a40f67d273f3167444a9479f859e70c50c.

benchmark/multidimensional_memset_benchmark_kernels.cpp
brycelelbach: Changed this to `ints` and changed `N` and `M` to `flux::distance_t` in 4132087e2b68de4ad69c0643e1cbaff2b67b5ecd.

benchmark/multidimensional_memset_benchmark_kernels.cpp
brycelelbach: This is fixed as of 4132087e2b68de4ad69c0643e1cbaff2b67b5ecd. The flux variant was iterating in the opposite order, leading to strided memory accesses that couldn't be vectorized on some platforms. Now that both the reference and the flux variant are doing the same thing, we get identical assembly.

include/flux/op/cartesian_product.hpp
```diff
@@ -88,24 +91,37 @@ struct cartesian_product_traits_base {
...
+        auto this_size = flux::size(base);
+
+        // If the new index overflows the maximum or underflows zero, calculate the carryover and fix it.
+        if (new_index < 0 || new_index >= this_size) {
+            offset = num::checked_div(new_index, this_size);
+            new_index = num::checked_mod(new_index, this_size);
+
+            // Correct for negative index which may happen when underflowing.
+            if (new_index < 0) {
+                new_index = num::checked_add(new_index, this_size);
```
tcbrindle, line 111: Massive pedantry: `this_size` must be positive (or zero), so if `new_index` is negative then this addition can never overflow and we don't technically need to do a `checked_add()`. But honestly it's probably not worth fixing.

include/flux/op/cartesian_product.hpp
```diff
@@ -216,6 +246,37 @@ struct sequence_traits<detail::cartesian_product_adaptor<Bases...>>
...
+            auto this_current = flux::for_each_while(std::get<I>(self.bases_),
+                [&](auto&& elem) {
+                    std::tie(keep_going, nested_current) = for_each_while_impl<I+1>(
+                        self, func, FLUX_FWD(partial_elements)..., FLUX_FWD(elem));
+                    return keep_going;
+                });
+            return std::tuple(keep_going,
+                std::tuple_cat(std::tuple(std::move(this_current)), std::move(nested_current)));
+        }
+    }
```
tcbrindle, line 278: Thinking out loud: rather than returning a `(keep_going, partial_cursor)` tuple, could `for_each_while_impl` take and pass a `bool&` out parameter instead, and just return the partial cursor? 

This would save having to do a `tuple_cat` in the return, which from memory is quite expensive compile-time wise, and might also save needing a local variable at every level?

tcbrindle (commented):
Numbers on my laptop:

With Clang 16:

| relative |               ns/op |                op/s |    err% |     total | benchmark
|---------:|--------------------:|--------------------:|--------:|----------:|:----------
|   100.0% |          637,413.32 |            1,568.84 |    5.8% |      0.30 | :wavy_dash: `memset_2d_reference` (Unstable with ~43.5 iters. Increase `minEpochIterations` to e.g. 435)
|    16.1% |        3,953,071.73 |              252.97 |    0.8% |      1.89 | `memset_2d_std_cartesian_product_iota`
|   117.2% |          543,979.24 |            1,838.31 |    1.0% |      0.26 | `memset_2d_flux_cartesian_product_iota`
|   100.0% |          434,128.10 |            2,303.47 |    2.2% |      0.21 | `memset_diagonal_2d_reference`
|    16.3% |        2,659,737.14 |              375.98 |    0.3% |      1.28 | `memset_diagonal_2d_std_cartesian_product_iota_filter`
|   100.3% |          432,654.76 |            2,311.31 |    1.9% |      0.21 | `memset_diagonal_2d_flux_cartesian_product_iota_filter`


With GCC 13.1:

| relative |               ns/op |                op/s |    err% |     total | benchmark
|---------:|--------------------:|--------------------:|--------:|----------:|:----------
|   100.0% |          910,021.74 |            1,098.87 |    5.9% |      0.44 | :wavy_dash: `memset_2d_reference` (Unstable with ~43.5 iters. Increase `minEpochIterations` to e.g. 435)
|   100.6% |          904,318.18 |            1,105.81 |    3.2% |      0.44 | `memset_2d_std_cartesian_product_iota`
|   105.0% |          866,658.54 |            1,153.86 |    1.7% |      0.41 | `memset_2d_flux_cartesian_product_iota`
|   100.0% |          585,775.00 |            1,707.14 |    1.7% |      0.28 | `memset_diagonal_2d_reference`
|    26.1% |        2,246,652.17 |              445.11 |    0.5% |      1.07 | `memset_diagonal_2d_std_cartesian_product_iota_filter`
|    34.0% |        1,723,046.51 |              580.37 |    1.0% |      0.83 | `memset_diagonal_2d_flux_cartesian_product_iota_filter`

It looks like GCC isn't able to vectorise the filter case, but Clang is doing great.

include/flux/op/cartesian_product.hpp
brycelelbach: I'd rather err on the side of caution and leave the check in. There are other places in the function where the checked numerics are important. I don't want future readers or modifiers of the function to have to discern which uses are significant and which are not.

include/flux/op/cartesian_product.hpp
brycelelbach: Done in abba492d018d7801c150117e4a320854a1a2b69b. I had to pass the cursor as a reference parameter as well. This removes the need for the partial cursor structure entirely.

include/flux/op/cartesian_product.hpp
```diff
@@ -216,6 +231,33 @@ struct sequence_traits<detail::cartesian_product_adaptor<Bases...>>
...
+        } else {
+            std::get<I>(cur) = flux::for_each_while(std::get<I>(self.bases_),
+                [&](auto&& elem) {
+                    for_each_while_impl<I+1>(
+                        self, keep_going, cur,
+                        func, FLUX_FWD(partial_elements)..., FLUX_FWD(elem));
+                    return keep_going;
+                });
+        }
+    }
```
tcbrindle, line 259: I love this implementation! It's super-clean, and nice to be able to avoid the `partial_cursor` work as well

include/flux/op/cartesian_product.hpp
```diff
@@ -242,6 +284,16 @@ struct sequence_traits<detail::cartesian_product_adaptor<Bases...>>
     {
         return read_(flux::move_at_unchecked, self, cur);
     }
+
+    template <typename Self, typename Function>
+    static constexpr auto for_each_while(Self& self, Function&& func)
+        -> cursor_t<Self>
+    {
+        bool keep_going = true;
+        cursor_t<Self> cur;
```
tcbrindle, line 293: Now whenever anybody asks me why cursors for multipass sequences need to be default-constructible, I'll have a good answer :)

tcbrindle (approved):

include/flux/op/cartesian_product.hpp
brycelelbach: Thanks! This is definitely much better, thanks for pushing me in the right direction.

include/flux/op/cartesian_product.hpp
brycelelbach: Ah, I just assumed they were!

PR #93, opened 2023-07-11
Compiler flag improvements

brycelelbach:
Turn on warnings and fix some subsequent issues, add a bunch of compiler diagnostic flags, tweak build mode flags.


tcbrindle:
Thanks for this Bryce! My CMake-fu is pretty weak so thanks for having a go at improving it

A couple of quick questions:

 * Can the `flux_docs` library be defined in `examples/CMakeLists.txt` instead, to keep things tidier in the top-level file? (Are there many examples which are causing unused variable warnings?)
 * I'm a little bit concerned about making decisions regarding optimisation settings (`-fomit-frame-pointer` etc) on the top-level `flux` library that would propagate to downstream users who might necessarily not want them... would it be better to add e.g. an empty `flux_internal` interface library that we could set all the optimisation flags for, and then have the test/benchmark executables use that?

tcbrindle:
@brycelelbach I'm going to close this as it's been a while and there have been quite a lot of CMake changes since. Feel free to submit another PR if you're still interested :)

Issue #99, opened 2023-07-13
Expanding the internal iteration API


tcbrindle:
Currently internal iteration is achieved by a sequence customisation point `for_each_while(predicate) -> cursor_t`, which instructs the sequence to iterate over all of its elements until such time as the predicate returns false.

Unfortunately there are some cases where we can't use internal iteration even though we'd like to. Algorithms like `fold_first` (#98), `min` and `max` need to inspect the initial element before iterating over the rest of the sequence; more generally, any time we take a slice of a sequence (such as with the elements of `chunk`, `chunk_by`, `slide` etc) then we can't implement `for_each_while()` on the slice in terms of an equivalent call on the base sequence, and so have to fall back to external iteration.

A solution to this would be to replace the existing `for_each_while()` customisation point with a new pair of functions taking extra cursor arguments to indicate where iteration should begin and end:

```cpp
auto iterate_while(auto& seq, auto&& pred, cursor_t from) -> cursor_t;
auto iterate_while(auto& seq, auto&& pred, cursor_t from, cursor_t to) -> cursor_t;
```

The slightly odd parameter order is to allow some sequence impls to provide just a single overload, with the `to` parameter defaulted to `flux::last()`. Implementations wouldn't need to provide both of these functions: in particular, most single-pass sequences aren't going to be able to provide the second one.

With this approach, algorithms like `fold_first` could examine the first element before proceeding with a call to `iterate_while()`; and `iterate_while()` on slices could be implemented in terms of calls to the equivalent function on their parent sequence. The existing `flux::for_each_while(seq, pred)` wrapper function could be implemented as a call to `iterate_while(seq, pred, first(seq))`.

This would add (yet) another customisation point that authors would have to worry about when writing a sequence implementation, which I'm somewhat reluctant to do, but I think the codegen benefits of internal iteration are great enough that it's justified in this case.

@brycelelbach @brevzin @jnytra  any thoughts?

brevzin:
Yeah I ran into the same problem when I was trying to do this, and instead came up with a janky semi-stateful internal iteration solution? Not amazing.

So I guess the implementation of `min` would look something like... this?

```cpp
template <sequence Seq>
auto min(Seq& seq) -> Optional<value_t<Seq>> {
    auto cursor = first(seq);
    if (is_last(seq, cursor)) {
        return {};
    }

    Optional<value_t<Seq>> best = read_at(seq, cursor);
    inc(seq, cursor);
[...]

PR #100, opened 2023-07-14
Update the internal iteration protocol

tcbrindle:
This is an experimental PR to try out the ideas in #99, namely replacing the existing `for_each_while(seq, pred) -> cursor_t` customisation point with a new pair of customisation points:

```cpp
auto iterate_while(auto& seq, auto pred, cursor_t from) -> cursor_t;
auto iterate_while(auto& seq, auto pred, cursor_t from, cursor_t to) -> cursor_t;
```

The idea being that this will allow us to use internal iteration in more places.

A rough plan of action:

* [x] Add `flux::iterate_while` wrapper functions
* [x] Implement `flux::for_each_while` using `iterate_while`
* [ ] Replace existing `for_each_while` custom implementations with `iterate_while` customisations:
    - [x] `T[N]`
    - [x] `std::reference_wrapper`
    - [x] `ranges::contiguous_range`
    - [ ]  `simple_sequence_base`
    - [ ] `chain_adaptor`
    - [ ] `cycle_adaptor`
    - [x] `filter_adaptor`
    - [ ] (multipass) `flatten_adaptor` 
    - [x] `map_adaptor`
    - [ ] `passthrough_traits_base`
    - [ ] `reverse_adaptor`
    - [ ] `scan_adaptor`
    - [ ] `scan_first_adaptor`
    - [ ] `take_adaptor`
    - [ ] `take_while_adaptor`
    - [ ] `array_ptr`
    - [ ] `range_sequence`
    - [ ] `repeat_sequence`
    - [ ] `single_sequence`
    - [ ] `unfold_sequence`
* [ ] Verify that we're again getting internal iteration happening in all the places it did before
* [ ] Update `fold_first` to use `iterate_while`
* [ ] Implement `iterate_while` for `subsequence`
* [ ] Add a benchmark to verify that we're now getting improved performance with the new changes
* [ ] Add documentation for `iterate_while`

### DIFF
--- a/include/flux/core/default_impls.hpp
+++ b/include/flux/core/default_impls.hpp
@@ -65,16 +65,16 @@ struct sequence_traits<T[N]> {
 
     static constexpr auto size(auto const&) -> distance_t { return N; }
 
-    static constexpr auto for_each_while(auto& self, auto&& pred) -> index_t
+    static constexpr auto iterate_while(auto& self, index_t from, index_t to,
+                                        auto&& pred) -> index_t
     {
-        index_t idx = 0;
-        while (idx < N) {
-            if (!std::invoke(pred, self[idx])) {
+        while (from <  to) {
+            if (!std::invoke(pred, self[from])) {
                 break;
             }
-            ++idx;
+            ++from;
         }
-        return idx;
+        return from;
     }
 };
 
@@ -163,6 +163,20 @@ struct sequence_traits<std::reference_wrapper<Seq>> {
     {
         return flux::move_at(self.get(), cur);
     }
+
+    static constexpr auto iterate_while(self_t self, cursor_t<Seq> from,
+                                        auto&& pred) -> cursor_t<Seq>
+    {
+        return flux::iterate_while(self.get(), std::move(from), FLUX_FWD(pred));
+    }
+
+    static constexpr auto iterate_while(self_t self, cursor_t<Seq> from,
+                                        cursor_t<Seq> to, auto&& pred)
+        -> cursor_t<Seq>
+    {
+        return flux::iterate_while(self.get(), std::move(from),
+                                   std::move(to), FLUX_FWD(pred));
+    }
 };
 
 // Default implementation for contiguous, sized ranges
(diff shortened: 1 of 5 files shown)


tcbrindle:
@brycelelbach Don't worry, I'm not going to change anything out from under you: even if I get this all done and it shows an improvement, I'll wait till you've finished your work on #92 and then update `cartesian_product` to use the new protocol  if necessary

brycelelbach:
@tcbrindle sounds good - I'll try to get you an updated PR sometime next week.

brycelelbach:
@tcbrindle On the topic of improving the internal iteration protocol, I'd like to see the protocol evolved to better support parallelism in the future. I'd love to use flux to prototype a next generation parallel algorithms library.

The customization point today, `for_each_while`, is an early-exit `for` loop abstraction.

In the non-parallel case, we can implement non-early-exit algorithms like `for_each` in terms of `for_each_while` with essentially zero cost. The inverse is not true; we can't implement early-exit algorithms like `find` in terms of a non-early-exit `for_each`.

However, in parallel, implementing non-early-exit algorithms like `for_each` in terms of `for_each_while` is NOT zero cost. To parallelize, you'd have to have some sort of concurrent rollback or cancellation mechanism. On some platforms, like GPUs, it's actually more efficient to implement parallel early-exit algorithms like `find` by NOT actually exiting early.

Some options:
* Have two customization points, `for_each` and `for_each_while`.
* Add a compile time parameter to `for_each_while` that indicates whether or not early-exit is needed.

tcbrindle:
> I'd love to use flux to prototype a next generation parallel algorithms library.

That would be awesome!

> Some options:
> * Have two customization points, for_each and for_each_while.
> * Add a compile time parameter to for_each_while that indicates whether or not early-exit is needed.

I guess in general we'd want some way to indicate that we want parallel execution (for all algorithms where it's possible), perhaps something like the standard execution policies? In that case we could say that `for_each_while` (or `iterate_while`), when passed a non-sequential execution policy, doesn't do early termination. Would that be sufficient?

tcbrindle:
This has bitrotted, and I have better ideas for how to solve the problem

Issue #57, opened 2023-05-01
Use low-level contiguous_sequence optimisations where appropriate


tcbrindle:
Currently we [optimise `output_to()`](https://github.com/tcbrindle/flux/blob/main/include/flux/op/output_to.hpp#L35) for `contiguous_sequence`s of trivial types to use `std::memmove`.

There are more places where we could use equivalent low-level optimisations, including

 * [ ] `fill()` (using `memset()`)
 * [ ] `equal()` (using `memcmp()`)
 * [ ] `compare()` (using `memcmp()`)
 * [ ] `find()` (using `memchr()`)
 * [x] ~~`search()` (using `memmem()` where available)~~

See also the [equivalent bug for Nanorange](https://github.com/tcbrindle/NanoRange/issues/107) with more information about when these optimisations can be used.

DeveloperPaul123:
I've started work on this in #103

DeveloperPaul123:
@tcbrindle I have a question while working on `equal()`

There is this test case:

```cpp
 // Different but comparable element types
{
    int arr1[] = {1, 2, 3, 4, 5};
    float arr2[] = {1.f, 2.f, 3.f, 4.f, 5.f};

    STATIC_CHECK(flux::equal(arr1, arr2));
}
```

It seems that the intension is to compare values and accept them as equal. If we want to optimize with `std::memcmp`, does it seem reasonable to require the the values of the 2 sequences be of the same type? Currently, using the `std::memcmp` optimization causes the above test to fail (where it previously passed).

DeveloperPaul123:
@tcbrindle Currently, `std::memmem` is a GNU extension. Is it desirable to use this on platforms that's it's available on? Currently C++ extensions are explicitly turned off via CMake

tcbrindle:
@DeveloperPaul123 Thinking about it some more, I think it's probably better to not use `memmem` for now and stick with portable standard library facilities... we can look at platform/vendor-specific optimisations later if/when they're needed.

tcbrindle:
Added in #103, thanks @DeveloperPaul123!

PR #103, opened 2023-07-25
Low level optimizations for contiguous sequences

DeveloperPaul123:
## Description

Adds low level optimizations to address #57 

**Optimizations**

- [x] `std::memset` with `fill()`
- [x] `std::memcmp` with `equal()`
- [x] `std::memcmp` with `compare()`
- [x] `std::memchr` with `find()`
- ~~`std::memmem` with `search()`~~

### DIFF
--- a/include/flux/core/utils.hpp
+++ b/include/flux/core/utils.hpp
@@ -70,6 +70,13 @@ FLUX_EXPORT
 template <std::integral To>
 inline constexpr auto checked_cast = detail::checked_cast_fn<To>{};
 
+namespace detail {
+
+template <typename T, typename... U>
+concept any_of = (std::same_as<T, U> || ...);
+
+} // namespace detail
+
 } // namespace flux
 
 #endif
--- a/include/flux/op/fill.hpp
+++ b/include/flux/op/fill.hpp
@@ -28,8 +28,9 @@ public:
         constexpr bool can_memset = 
             contiguous_sequence<Seq> &&
             sized_sequence<Seq> &&
+            std::same_as<Value, value_t<Seq>> &&
             // only allow memset on single byte types
-            sizeof(value) == sizeof(std::uint8_t) &&
+            sizeof(value_t<Seq>) == 1 &&
             std::is_trivially_copyable_v<value_t<Seq>>;
 
         if constexpr (can_memset) {
--- a/include/flux/op/fill.hpp
+++ b/include/flux/op/fill.hpp
@@ -38,14 +38,14 @@ public:
                 impl(seq, value);
             } else {
                 auto size = flux::usize(seq);
-                if(size == 0){
+                if(size == 0) {
                     return;
                 }
                 
                 FLUX_ASSERT(flux::data(seq) != nullptr);
                 
                 std::memset(flux::data(seq), value,
-                    flux::usize(seq) * sizeof(value_t<Seq>));
+                    size * sizeof(value_t<Seq>));
             }
         } else {
             impl(seq, value);
--- a/include/flux/op/fill.hpp
+++ b/include/flux/op/fill.hpp
@@ -6,18 +6,42 @@
 #define FLUX_OP_FILL_HPP_INCLUDED
 
 #include <flux/op/for_each.hpp>
+#include <type_traits>
+#include <cstring>
 
 namespace flux {
 
 namespace detail {
 
 struct fill_fn {
+private:
+    template <typename Seq, typename Value>
+    static constexpr auto impl(Seq& seq, Value& value)
+    {
+        flux::for_each(seq, [&value](auto&& elem) { FLUX_FWD(elem) = value; });
+    }
+
+public:
     template <typename Value, writable_sequence_of<Value> Seq>
     constexpr void operator()(Seq&& seq, Value const& value) const
     {
-        flux::for_each(seq, [&value](auto&& elem) {
-            FLUX_FWD(elem) = value;
-        });
+        constexpr bool can_memset = 
+            contiguous_sequence<Seq> &&
+            sized_sequence<Seq> &&
+            // only allow memset on single byte types
+            sizeof(value) == sizeof(std::uint8_t) &&
+            std::is_trivially_copyable_v<value_t<Seq>>;
+
+        if constexpr (can_memset) {
+            if (std::is_constant_evaluated()) {
+                impl(seq, value);
+            } else {
+                std::memset(flux::data(seq), value,
+                    flux::usize(seq) * sizeof(value_t<Seq>));
+            }
+        } else {
+            impl(seq, value);
+        }
     }
 };
 
--- a/include/flux/op/fill.hpp
+++ b/include/flux/op/fill.hpp
@@ -37,6 +37,9 @@ public:
             if (std::is_constant_evaluated()) {
                 impl(seq, value);
             } else {
+                auto size = flux::usize(seq);
+                if(size == 0) return;
+                
                 FLUX_ASSERT(flux::data(seq) != nullptr);
                 
                 std::memset(flux::data(seq), value,
--- a/include/flux/op/fill.hpp
+++ b/include/flux/op/fill.hpp
@@ -37,6 +37,8 @@ public:
             if (std::is_constant_evaluated()) {
                 impl(seq, value);
             } else {
+                FLUX_ASSERT(flux::data(seq) != nullptr);
+                
                 std::memset(flux::data(seq), value,
                     flux::usize(seq) * sizeof(value_t<Seq>));
             }
--- a/include/flux/op/fill.hpp
+++ b/include/flux/op/fill.hpp
@@ -38,7 +38,9 @@ public:
                 impl(seq, value);
             } else {
                 auto size = flux::usize(seq);
-                if(size == 0) return;
+                if(size == 0){
+                    return;
+                }
                 
                 FLUX_ASSERT(flux::data(seq) != nullptr);
                 
--- a/include/flux/op/fill.hpp
+++ b/include/flux/op/fill.hpp
@@ -16,7 +16,7 @@ namespace detail {
 struct fill_fn {
 private:
     template <typename Seq, typename Value>
-    static constexpr auto impl(Seq& seq, Value& value)
+    static constexpr auto impl(Seq& seq, Value const& value)
     {
         flux::for_each(seq, [&value](auto&& elem) { FLUX_FWD(elem) = value; });
     }
--- a/include/flux/op/find.hpp
+++ b/include/flux/op/find.hpp
@@ -43,7 +43,7 @@ public:
             auto location = std::memchr(flux::data(seq), static_cast<unsigned char>(value),
                 flux::usize(seq) * sizeof(value_t<Seq>));
             if (location == nullptr) {
-                return flux::usize(seq);
+                return flux::last(seq);
             } else {
                 auto offset = static_cast<value_t<Seq> const*>(location) - flux::data(seq);
                 return flux::next(seq, flux::first(seq), offset);
--- a/include/flux/op/find.hpp
+++ b/include/flux/op/find.hpp
@@ -41,12 +41,12 @@ public:
                 return impl(seq, value);
             } else {
                 auto size = flux::usize(seq);
-                if(size == 0) {
+                if (size == 0) {
                     return flux::last(seq);
                 }
                 FLUX_ASSERT(flux::data(seq) != nullptr);
                 auto location = std::memchr(flux::data(seq), static_cast<unsigned char>(value),
-                    flux::usize(seq) * sizeof(value_t<Seq>));
+                    size * sizeof(value_t<Seq>));
                 if (location == nullptr) {
                     return flux::last(seq);
                 } else {
--- a/include/flux/op/find.hpp
+++ b/include/flux/op/find.hpp
@@ -39,16 +39,17 @@ public:
         if constexpr (can_memchr) {
             if (std::is_constant_evaluated()) {
                 return impl(seq, value);
-            }
-            auto location = std::memchr(flux::data(seq), static_cast<unsigned char>(value),
-                flux::usize(seq) * sizeof(value_t<Seq>));
-            if (location == nullptr) {
-                return flux::last(seq);
             } else {
-                auto offset = static_cast<value_t<Seq> const*>(location) - flux::data(seq);
-                return flux::next(seq, flux::first(seq), offset);
+                FLUX_ASSERT(flux::data(seq) != nullptr);
+                auto location = std::memchr(flux::data(seq), static_cast<unsigned char>(value),
+                    flux::usize(seq) * sizeof(value_t<Seq>));
+                if (location == nullptr) {
+                    return flux::last(seq);
+                } else {
+                    auto offset = static_cast<value_t<Seq> const*>(location) - flux::data(seq);
+                    return flux::next(seq, flux::first(seq), offset);
+                }
             }
-
         } else {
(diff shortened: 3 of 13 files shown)


DeveloperPaul123:
@tcbrindle I filled this PR to see if I'm on the right track (any feedback much appreciated) and also to make it know that the work is in progress for that particular issue.

tcbrindle:
Thanks for working on this! It looks like there's a syntax error which is tripping up the non-Windows builds, but the general approach looks good to me 👍

include/flux/core/utils.hpp
```diff
@@ -70,6 +68,36 @@ FLUX_EXPORT
...
+};
+
+template <class... Traits>
+inline constexpr bool disjunction_v = disjunction<Traits...>::value;
+
+// VARIABLE TEMPLATE _Is_any_of_v
+template <class T, class... Types>
+inline constexpr bool is_any_of = disjunction_v<std::is_same<T, Types>...>;
+;
+} // namespace detail
```
tcbrindle, line 100: FYI, there's a `detail::any_of` concept in `split_string.hpp` that is a bit simpler than the C++17 approach :). You can move it to utils.hpp if you like. 

https://github.com/tcbrindle/flux/blob/main/include/flux/op/split_string.hpp#L17-L18

tcbrindle: (Not a full review as I can see you're still working on it, I just happened to notice this :) )

DeveloperPaul123: Oh thank you! That's very helpful :)

DeveloperPaul123:
I've noticed that with the `any_of` move, `split()` seems to have broken? I'm getting an exception with this code in `test_split.cpp`

```cpp
{
    auto sv = "the quick brown fox"sv;

    auto split = flux::split(flux::ref(sv), ' ');

    using S = decltype(split);

    static_assert(flux::multipass_sequence<S>);
    static_assert(flux::contiguous_sequence<flux::element_t<S>>);

    static_assert(flux::multipass_sequence<S const>);
    static_assert(flux::contiguous_sequence<flux::element_t<S const>>);

    STATIC_CHECK(check_equal(std::move(split).map(to_string_view),
            std::array{"the"sv, "quick"sv, "brown"sv, "fox"sv}));
}
```

DeveloperPaul123:
Just looking for an initial round of feedback; I'll have to address the build issues.

include/flux/op/find.hpp
```diff
@@ -6,44 +6,71 @@
...
+    constexpr auto operator()(Seq&& seq, Value const& value) const
+        -> cursor_t<Seq>
     {
-        return for_each_while(seq, [&](auto&& elem) {
-            return FLUX_FWD(elem) != value;
-        });
+        constexpr auto memchr_opt = flux::contiguous_sequence<Seq> &&
+            flux::sized_sequence<Seq> && std::is_integral_v<Value> &&
+            flux::detail::any_of<value_t<Seq>, char, signed char, unsigned char,
+                const char, const signed char, const unsigned char>;
```
tcbrindle, line 37: This should be `can_memchr` for consistency with the other functions. 

I'm not quite sure that I understand the MSVC condition here allowing `Value` to be any integral type -- don't we need to also have `sizeof(Value) == 1` since `memchr` looks for a single byte? But maybe I'm missing something?

Conservatively (at least until I get my head around the above!) I think we should require that `Value` is the same type as `value_t<Seq>`, which is one of the "known good" types -- but i think we can also expand that list to include C++20 `char8_t` and `std::byte`. (We don't need to list the const variants as `value_t` shouldn't be const-qualified.)

```suggestion
        constexpr auto can_memchr = 
            contiguous_sequence<Seq> && sized_sequence<Seq> && 
            std::same_as<Value, value_t<Seq>> &&
            detail::any_of<value_t<Seq>, char, signed char, unsigned char, char8_t, std::byte>;
```

include/flux/op/find.hpp
```diff
@@ -6,44 +6,71 @@
...
+            if (std::is_constant_evaluated()) {
+                return impl(seq, value);
+            }
+            auto location = std::memchr(flux::data(seq), value,
+                flux::usize(seq) * sizeof(value_t<Seq>));
+            if (location == nullptr)
+                return flux::usize(seq);
+            else
+                return static_cast<const value_t<Seq>*>(location) -
+                    flux::data(seq);
```
tcbrindle, line 49: Braces on if statements, please :)

This isn't quite returning the right thing. `flux::find` returns a *cursor*, which for a contiguous sequences is an integer index -- but for subsequences, the first element might not have index zero, so we have to adjust the return value. This is what is causing the test failures with `split`, because it calls `find()` on a subsequence, hits the new specialisation and gets the wrong answer.

```suggestion
            if (location == nullptr) {
                return flux::last(seq);
            } else {
                auto offset = static_cast<value_t<Seq> const*>(location) - flux::data(seq);
                return flux::next(seq, flux::first(seq), offset);
            }
```

tcbrindle (changes requested):
There are quite a few comments, but several of them are just minor style things. 

This is looking in pretty good shape, we just need to make sure that the conditions for using the specialisations is correct in each case. I'd rather err on the side of caution -- it's better to use a potentially slower code path than give the wrong answer or (worse) end up with UB.

There do seem to be a lot of unnecessary, unrelated formatting changes though -- I guess from clang-format messing things up? I started marking them but gave up as there are quite a few... Please put these lines back to how they were originally so that the diff only includes meaningful changes.

tcbrindle:
> I've noticed that with the `any_of` move, `split()` seems to have broken? I'm getting an exception with this code in `test_split.cpp`
> 
> ```c++
> {
>     auto sv = "the quick brown fox"sv;
> 
>     auto split = flux::split(flux::ref(sv), ' ');
> 
>     using S = decltype(split);
> 
>     static_assert(flux::multipass_sequence<S>);
>     static_assert(flux::contiguous_sequence<flux::element_t<S>>);
> 
>     static_assert(flux::multipass_sequence<S const>);
>     static_assert(flux::contiguous_sequence<flux::element_t<S const>>);
> 
>     STATIC_CHECK(check_equal(std::move(split).map(to_string_view),
>             std::array{"the"sv, "quick"sv, "brown"sv, "fox"sv}));
> }
> ```

This is caused by the new specialisation of `find` (which `split` uses) returning the wrong thing in some cases. See the review comments above (the `any_of` thing is just a coincidence).

include/flux/op/find.hpp
DeveloperPaul123: > I'm not quite sure that I understand the MSVC condition here allowing Value to be any integral type

I did this because I saw that the the search value type in the function signature of `std::memchr` is `int` but I realize now it's meant to be a byte. 🤦‍♂️

include/flux/op/find.hpp
DeveloperPaul123: Thanks for this! I made the changes you suggested but I'm still getting a failure on the unit tests. I'll have to dig in deeper.

DeveloperPaul123: Disregard, that was my mistake.

DeveloperPaul123:
I believe I've addressed most, if not all of your comments. Unit tests are also all passing for me. Please take another look when you have a moment. Thanks!

include/flux/op/fill.hpp
```diff
@@ -6,18 +6,42 @@
...
     {
-        flux::for_each(seq, [&value](auto&& elem) {
-            FLUX_FWD(elem) = value;
-        });
+        constexpr bool can_memset = 
+            contiguous_sequence<Seq> &&
+            sized_sequence<Seq> &&
+            // only allow memset on single byte types
+            sizeof(value) == sizeof(std::uint8_t) &&
+            std::is_trivially_copyable_v<value_t<Seq>>;
```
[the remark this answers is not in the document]
tcbrindle: I think this change is still required

tcbrindle (changes requested):
This is looking really good! There's only one change still outstanding, namely the definition of `can_memset` in `fill.hpp`

DeveloperPaul123:
Whoops! I missed that, thanks!

tcbrindle (dismissed):
Thanks for addressing all the feedback, this looks great now!

DeveloperPaul123:
Great!

Is there a need for more unit tests to make codecov happy?

tcbrindle:
Looking at the CodeCov complaints, most of them seem to be  because it doesn't understand `std::is_constant_evaluated()` This is technically a runtime branch, but we can't ever test it in a way that will make CodeCov happy so we'll just ignore it.

The other complaint is that we're not testing the short-circuit case in `equal` when both sequences are sized but their sizes differ... but that [definitely looks like it's being tested](https://github.com/tcbrindle/flux/blob/7d6301fb928cca9cea2a55387b9d36d77e5265ea/test/test_equal.cpp#L51-L58) so my guess is that the compiler has inlined the it into the test function, so as far as CodeCov is concerned the line is never executed.

So I think we're fine :)

DeveloperPaul123:
Done! No problem at all. But note that tests now fail. Do we need to take into account `flux::empty<>` sequences? 

```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test-libflux.exe is a Catch v2.13.7 host application.
Run with -? for options

-------------------------------------------------------------------------------
drop
-------------------------------------------------------------------------------
flux/test/test_drop.cpp(104)
...............................................................................

flux/test/test_drop.cpp(104): FAILED:
due to unexpected exception with message:
  flux/include\flux/op/equal.hpp:62: Fatal error: assertion    
  'flux::data(seq2) != nullptr' failed

===============================================================================
test cases:  74 |  73 passed | 1 failed
assertions: 556 | 555 passed | 1 failed
```

tcbrindle:
> Done! No problem at all. But note that tests now fail. Do we need to take into account flux::empty<> sequences?

Thanks! I'd completely forgotten about `flux::empty`... so much for "nullptr should never happen"! (But it's good that we caught it)

So I guess what we really need to do is check the size of the sequence first... if that is zero then we can exit early, as there's no work to do. Otherwise, we get the data pointer and assert that it's not nullptr, and then finally call the C library function. That is, something like:

```cpp
auto size = flux::usize(seq);
if (size == 0) {
    return;
}
auto data = flux::data(seq);
FLUX_ASSERT(data != nullptr);

memset(data, val, size); 
```

(adjusted as appropriate for each case).

This turned out to be more complicated than I thought!

DeveloperPaul123:
Sounds good, I'll get to work on that. 

How would you like to handle the different cases in `compare()`? What would make the most sense if both sequences are empty? Or if only one sequence is empty?

Maybe I should just look at the unit tests 😅

tcbrindle:
> How would you like to handle the different cases in `compare()`? What would make the most sense if both sequences are empty? Or if only one sequence is empty?

For `compare()`, I think we get the right behaviour if we do something like:

```cpp
auto const seq1_size = flux::usize(seq1);
auto const seq2_size = flux::usize(seq2);
auto min_size = std::min(seq1_size, seq2_size);

int cmp_result = 0;

if (min_size > 0) {
    auto const data1 = flux::data(seq1);
    FLUX_ASSERT(data1 != nullptr);
    auto const data2 = flux::data(seq2);
    FLUX_ASSERT(data2 != nullptr);

    cmp_result = std::memcmp(data1, data2, min_size);
}
```

That is, we only call the C function if the number of bytes is greater than zero, having first asserted that the data pointer is not nullptr.

This is getting messier and messier, sorry!

DeveloperPaul123:
Ok I think this is ready now and all tests pass for me!

include/flux/op/fill.hpp
```diff
@@ -6,18 +6,50 @@
...
+            // only allow memset on single byte types
+            sizeof(value_t<Seq>) == 1 &&
+            std::is_trivially_copyable_v<value_t<Seq>>;
+
+        if constexpr (can_memset) {
+            if (std::is_constant_evaluated()) {
+                impl(seq, value);
+            } else {
+                auto size = flux::usize(seq);
+                if(size == 0){
```
tcbrindle, line 41: Style nitpick

```suggestion
                if (size == 0) {
```

include/flux/op/fill.hpp
```diff
@@ -6,18 +6,50 @@
...
+                    return;
+                }
+                
+                FLUX_ASSERT(flux::data(seq) != nullptr);
+                
+                std::memset(flux::data(seq), value,
+                    flux::usize(seq) * sizeof(value_t<Seq>));
```
tcbrindle, line 48: We've already got the size, we may as well re-use it

```suggestion
                std::memset(flux::data(seq), value,
                    size * sizeof(value_t<Seq>));
```

include/flux/op/find.hpp
```diff
@@ -6,21 +6,58 @@
...
+            contiguous_sequence<Seq> && sized_sequence<Seq> && 
+            std::same_as<Value, value_t<Seq>> &&
+            flux::detail::any_of<value_t<Seq>, char, signed char, unsigned char, char8_t, std::byte>;
+
+        if constexpr (can_memchr) {
+            if (std::is_constant_evaluated()) {
+                return impl(seq, value);
+            } else {
+                auto size = flux::usize(seq);
+                if(size == 0) {
```
tcbrindle, line 44: Style nitpick

```suggestion
                if (size == 0) {
```

include/flux/op/find.hpp
```diff
@@ -6,21 +6,58 @@
...
+                    return flux::last(seq);
+                }
+                FLUX_ASSERT(flux::data(seq) != nullptr);
+                auto location = std::memchr(flux::data(seq), static_cast<unsigned char>(value),
+                    flux::usize(seq) * sizeof(value_t<Seq>));
```
tcbrindle, line 49: Again, we can re-use the `size` variable

```suggestion
                auto location = std::memchr(flux::data(seq), static_cast<unsigned char>(value),
                    size * sizeof(value_t<Seq>));
```

single_include/flux.hpp
tcbrindle, line 1: This file is automatically generated, so modifications shouldn't be committed

tcbrindle (changes requested):
We can simplify the test in equal(), because we already know that both sequences are the same size. Other than that, just a couple of tiny code style things.

CodeCov is complaining that the new "size == 0" code-paths aren't being tested, so it's probably worth adding a couple of little tests to cover this -- but I'm happy to do it if you've had enough of this PR already!

Also, it looks like the new compare() specialisation isn't actually reached by any of the existing tests... This definitely needs rectifying before we can commit this, but again, I'm happy to add some unsigned char compare() tests if you don't fancy it.

DeveloperPaul123:
I was able to add some tests, but not sure if I covered everything you had in mind. I also address all (I think) your feedback from the latest review. Sorry for the silly oversights at times and thanks for all the feedback!

DeveloperPaul123:
Looks like I didn't cover all the cases for `compare()`. Some help with this would be appreciated.

tcbrindle (changes requested):
All the actual implementation code looks great now!

`compare()` is still missing a couple of test cases, I've made some suggestions below.

CodeCov is also complaining that we're not testing the case of two zero-length contiguous sequences in `equal()`. Something like

```cpp
std::array<int, 0> arr;
STATIC_CHECK(flux::equal(arr, arr) == true);
```

in test_equal.cpp should keep it happy I think?

DeveloperPaul123:
Done!

tcbrindle (approved):

tcbrindle:
It's done!! 🎉 

Thanks so much!

PR #109, opened 2023-08-01
More modules tinkering

tcbrindle:
* Inline some non-template member functions
* Ensure the module can be built/imported in Windows CI runs


tcbrindle:
Alas, trying to build the module in Clang/MacOS CI runs into difficulties. I'll try again another time.

PR #118, opened 2023-08-10
Re-add read_only_adaptor::move_at specialisation

### LINKED ISSUE #117: Static assertion build error in Visual Studio 17.7
When using the Visual Studio generator to build `flux` I'm now getting this build error:

```
flux\test\test_read_only.cpp(132,28): error C2607: static assertion failed
flux\test\test_read_only.cpp(132,28): message : the concept 'std::same_as<std::pair<const int &,const double &>,std::pair<const int &&,const double &&>>' evaluated to false
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.37.32822\include\concepts(36,9): message : 'std::pair<const int &,const double &>' and 'std::pair<const int &&,const double &&>' are different types
```

The offending code in `concepts.hpp`. `FLUX_HAVE_CPP23_TUPLE_COMMON_REF` is `true` in this context.

```cpp
#ifdef FLUX_HAVE_CPP23_TUPLE_COMMON_REF
    std::common_reference_with<element_t<Seq>&&, value_t<Seq>&> &&
    std::common_reference_with<rvalue_element_t<Seq>&&, value_t<Seq> const&> &&
#endif
```

tcbrindle:
Since #97 this was falling back to `map_adaptor::move_at()`, which does the wrong thing in C++23 mode when the base element type is something like `tuple<T&>`.

Fixes #117

### DIFF
--- a/include/flux/op/read_only.hpp
+++ b/include/flux/op/read_only.hpp
@@ -30,7 +30,24 @@ public:
     {}
 
     struct flux_sequence_traits : map::flux_sequence_traits {
+    private:
+        using const_rvalue_element_t = std::common_reference_t<
+            value_t<Base> const&&, rvalue_element_t<Base>>;
+
+    public:
         using value_type = value_t<Base>;
+
+        static constexpr auto move_at(auto& self, cursor_t<Base> const& cur)
+            -> const_rvalue_element_t
+        {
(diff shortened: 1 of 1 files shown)


tcbrindle:
@brevzin sorry for making `read_only_adaptor` a little less elegant again!

PR #136, opened 2023-11-27
Add Clang 17 on MacOS to CI

tcbrindle:
Our Mac CI script was mistakenly mixing up the Homebrew names "LLVM" and "LLVM@16", meaning it broken when Homebrew updated the default version.

We'll take this opportunity to not only fix it, but test with both Clang 16 and 17


tcbrindle:
Windows CI isn't working right now, but that #135

PR #144, opened 2023-12-13
Fix missing reverse_adaptor constraints

### LINKED ISSUE #143: Non const-iterable reversed sequences hard error when used as a range
Non const-iterable reversed sequences are pretty rare, but if you find one and try to use it in a range-for loop (or anything else that calls member `begin()`) it will hard error. 

This is because in checking the constraints for `begin() const`, it tries to check the return type of `sequence_traits<reverse_adaptor>::read_at` which returns `decltype(auto)`, meaning the compiler will instantiate the function to find out the return type. Unfortunately, the body of this function calls `flux::prev`, which then hard errors.

There are a couple of possible solutions, but the easiest is probably to avoid `decltype(auto)` and instead use `decltype(<actual expression>)` so we get graceful return type SFINAE.

This is pretty much the same as #47, but with `reverse_adaptor` rather than `zip_adaptor`.

tcbrindle:
Fixes #143

### DIFF
--- a/include/flux/op/reverse.hpp
+++ b/include/flux/op/reverse.hpp
@@ -63,25 +63,29 @@ public:
             return cur.base_cur == flux::first(self.base_);
         }
 
-        static constexpr auto read_at(auto& self, cursor_type const& cur)
+        template <typename Self>
+        static constexpr auto read_at(Self& self, cursor_type const& cur)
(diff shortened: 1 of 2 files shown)


tcbrindle:
Woo, ICEs in two different compilers at once, this must be a good commit...

PR #147, opened 2024-01-10
Migrate to Catch2 v3

tcbrindle:
Most of our tests are done at constexpr time, but we also use Catch2 to run the same tests at runtime.

Until now, we've bundled an old version of the Catch2 v2 header for this purpose. Meanwhile, Catch2 v3 has been released, which is no longer header-only, and is now a "proper" library.

This commit migrates us to using Catch2 v3, downloading it via CMake's "FetchContent" if it's not already available on the system. This has the benefit that it will be much easier to update Catch in the future (by just updating the commit id in tests/CMakeLists.txt), and also means that we no longer need to keep copies of Catch.cmake and CatchAddTests.cmake (as these can now be found via the downloaded library).

Unfortunately, the latest Catch includes a macro called STATIC_CHECK, which clashes with our own macro of the same name defined in test_utils.hpp. Because this macro is used literally thousands of times in our tests, this commit takes the easy route of #undef-ing the Catch version instead of changing our macro name.

The stated goal of Catch moving away from a single header is to improve compile times: however, since now we'll need to download and build the Catch library for every single CI run, I'm a bit worried our CI times might go through the roof. Let's see what happens...


tcbrindle:
...and the results are in. CI times are a little slower, but not dramatically so, and I think the trade-off of making it easier to keep up with the latest Catch development is probably worth it.

Issue #138, opened 2023-11-30
Add product/permutations/combinations from Python itertools


tcbrindle:
We currently have `cartestian_product(seq1, seq2, ...)` which takes a pack of N sequences and yields N-tuples of all the possible combinations of elements from those sequences. This is equivalent to a set of N nested `for` loops. This is equivalent to one form of the Python itertools [`product`](https://docs.python.org/3/library/itertools.html#itertools.product) operation.

It's pretty common however to want to use the *same* sequence repeatedly with `cartesian_product`. Itertools provides a second form of `product` for this, using a `repeat` argument. It would be good if we provided an equivalent too.

We'd need it to have a different name, and we'd probably want to take `N` as a template parameter to avoid having to allocate. So we could have something like `products<3>(seq)`, which would be equivalent to `cartestian_product(seq, seq, seq)` except that we wouldn't have to store multiple references to/copies of the sequence. The size of the resulting sequence would be `size(seq)^N`.

Itertools also provides the [`permutations`](https://docs.python.org/3/library/itertools.html#itertools.permutations) operation, which would also be nice to have. This yields (in lexicographical order) all the length-N permutations of the elements of the input sequence. This is a strict subset of the elements of `products<N>`, skipping those tuples which contain repeated elements from the underlying sequence. If `S` is the size of the original sequence, the size of `permutations<N>` is `S!/(N - S)!`.

Finally, Itertools has two slightly different [`combinations`](https://docs.python.org/3/library/itertools.html#itertools.combinations) operations. The first gives a strict subset of `permutations`, skipping tuples in which the elements are not in "sorted" order (according to their position in the initial sequence), and has size `S!/N!(N-S!)`

The second, the slightly oddly named (to me) [`combinations.with_replacement`](https://docs.python.org/3/library/itertools.html#itertools.combinations_with_replacement). This does the same as `combinations` except that it additionally allows individual elements to be repeated -- meaning the resulting elements are a subset of `product<N>`, but not `permutations<N>`. The number of elements is `(S+N-1)!/N!(S-1)!`. 

To make this a bit more concrete, imagine an input sequence`seq = [a, b, c, d]`. Then the various operations give:

`products<3>(seq)`: (4^3 = 64 elements)

```
aaa, aab, aac, aad, aba, abb, abc, abd, aca, acb, acc, acd, ada, adb, adc, add
[...]

PR #162, opened 2024-01-28
`cartesian_power`, `cartesian_power_map`

isaacy2012:
Attempts to partially address #138 regarding the `cartesian_power`, `cartesian_power_map` functionalities. 

Implementation centers around replacing `std::get<I>(self.bases_)` in `cartesian_product` with `self.base_` for `cartesian_power`.

Moves common logic across `cartesian_product`, `cartesian_power` and their `map` variants to a common base trait class `cartesian_base` with enum selectors for [product/power] and [tuple/map].


test/test_cartesian_power.cpp
```diff
@@ -0,0 +1,301 @@
...
+  constexpr auto operator()(Seq &&seq) const {
+    return impl(FLUX_FWD(seq), std::make_index_sequence<RepeatCount>{});
+  }
+};
+
+}
+
+template<int RepeatCount>
+FLUX_EXPORT inline constexpr auto cartesian_product_repeat_n = detail::cartesian_product_repeat_n_fn<RepeatCount>{};
+}
```
tcbrindle, line 46: This is really testing the existing `cartesian_product` rather than the new `cartesian_power`, so I don't think it's necessary here

test/test_cartesian_power.cpp
```diff
@@ -0,0 +1,301 @@
...
+      int sum_i = 0;
+      int sum_j = 0;
+      cart.for_each(flux::unpack([&] (int i, int j) {
+        sum_i += i;
+        sum_j += j;
+      }));
+      STATIC_CHECK(sum_i == 3 * (100 + 200 + 300));
+      STATIC_CHECK(sum_j == 3 * (100 + 200 + 300));
+
+    }
```
tcbrindle: See above, I don't think we need this

test/test_cartesian_power.cpp
tcbrindle, line 1: Suggestion: add a test for `cartesian_power<1>` as well, to make sure it returns the same elements as the original sequence (but wrapped in a  tuple)

test/test_cartesian_power_map.cpp
tcbrindle, line 1: As above, it would be nice to have a test for `cartesian_power_map<1>`, which I think should behave exactly like `flux::map`?

include/flux/op/cartesian_base.hpp
```diff
@@ -0,0 +1,359 @@
...
+
+        return cur;
+    }
+
+    template <std::size_t I, typename Self>
+    static constexpr auto ra_inc_impl(Self& self, cursor_t<Self>& cur, distance_t offset)
+    -> cursor_t<Self>&
+    {
+        if (offset == 0)
+            return cur;
```
tcbrindle, line 115: Style: braces after if statements please 
```suggestion
        if (offset == 0) {
            return cur;
        }
```

tcbrindle (changes requested):
Initial review: this is looking really good! 

I've suggested a few changes but the majority are stylistic, or because of other changes in Flux since I wrote the initial version of `cartesian_product`

include/flux/op/cartesian_base.hpp
```diff
@@ -0,0 +1,359 @@
...
+        }(self.base_, std::make_index_sequence<Arity>{});
+    }
+
+    template <typename Self>
+    requires (CartesianKind == cartesian_kind::product
+              && (sized_sequence<const_like_t<Self, Bases>> && ...))
+    static constexpr auto size(Self& self) -> distance_t
+    {
+        return std::apply([](auto&... base) {
+            return (flux::size(base) * ...);
```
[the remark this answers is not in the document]
isaacy2012: I've attempted to fix it here, but I feel as though I might be overcomplicating it: 

https://github.com/isaacy2012/flux/pull/1

include/flux/op/cartesian_base.hpp
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    template <typename Self>
+    static constexpr auto first(Self& self)
+        requires (CartesianKind == cartesian_kind::power)
+    {
+        return []<std::size_t... Is>(auto&& arg, std::index_sequence<Is...>) {
+            std::array<cursor_t<require_single_type_t<Bases...>>, Arity> cur = {(static_cast<void>(Is), flux::first(arg))...};
+            return cur;
+        }(self.base_, std::make_index_sequence<Arity>{});
```
tcbrindle, line 226: This is going to result in `Arity` calls to `flux::first(self.base_)`, which could result in a long chain of function calls if we're at the end of a long pipeline. But since we know `Base` is multipass, we know that its cursor type is copyable, so we can instead just call `flux::first(self.base_)` once and copy it into the array N times.

(Also, I don't think we need the `require_single_type_t` helper, because we know the `Bases...` pack only has one element in this function?)

```suggestion
        auto base_cur = flux::first(self.base_);
        return [&]<std::size_t... Is>(std::index_sequence<Is...>) {
            std::array<cursor_t<Bases>..., Arity> cur{(static_cast<void>(Is), base_cur)...};
            return cur;
        }(std::make_index_sequence<Arity>{});
```

tcbrindle (changes requested):
Thanks for making the changes, this is looking in really good shape! I've noted a couple more things (including addressing the CI failures) but they're all pretty minor.

After that, we just need to finalise the names of the new adaptors. I've been thinking about it and I think it's probably best to avoid using the word "power", as people might confuse it for a something which generates a [power set](https://en.wikipedia.org/wiki/Power_set).

 The only reasonable alternatives I've been able to think of are `cartesian_product_repeat<N>` and `cartesian_product_repeat_map<N>`, closely following the Python naming. I'd prefer something that was a bit shorter, but these are at least very explicit about what they do, so I think they're what we should go with -- unless you have any other suggestions?

include/flux/op/cartesian_base.hpp
isaacy2012: Ah yep, I made the `require_single_type_t` helper specifically to explicitly convey to a reader that we're asserting that the `Bases...` pack only has one element in the function - I thought it better to show intent there and protect against if it was ever accidentally changed. What are your thoughts?

tcbrindle: I like the intent, but I think this is probably better conveyed by saying `static_assert(sizeof...(Bases) == 1)`, which avoids the extra template instantiations that would be caused by a `require_single_type` class template (and actually lead to a better error message if this does gets screwed up at some point in the future).

tcbrindle:
> > unless you have any other suggestions?
> 
> Not really, but I got `cartesian_power` from this wikipedia page: https://en.wikipedia.org/wiki/Cartesian_product
> 
> I've changed it back to `cartesian_product_repeat` though, as if people are used to the Python way then it's probably better than the technically correct but potentially confusing `power`.

Hmm, I originally just googled the words `cartesian power` and it returned lots of results about cartesian products and power sets, which is why I was initially reluctant.

But now I've just searched again, using `"cartesian power"` (with quotes) and I see that actually the term is used in a few places -- for example [Microsoft's Azure Quantum language](https://learn.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.canon.iteratethroughcartesianpower), as well as some sites explaining mathematical terminology.

So if it's the right term mathematically, and it has precedent in other places, and it has a nice definition we can point to in the documentation.... then I guess `cartesian_power` *is* what we should call it after all?

I feel bad asking you to change it back again! But I see that it was all in the most recent commit so hopefully it's easy to revert. Sorry for getting you to do unnecessary work, but you've convinced me that you were right all along :)

include/flux/op/cartesian_base.hpp
isaacy2012: I see, I hadn't considered the cost of extra template instantiations. I've changed it to the static assert 👍

isaacy2012:
> I feel bad asking you to change it back again! But I see that it was all in the most recent commit so hopefully it's easy to revert. Sorry for getting you to do unnecessary work, but you've convinced me that you were right all along :)

No worries, I've renamed the branch to reflect the finalised name, so I'll make the updated PR from that branch.

PR #163, opened 2024-01-29
Feat/cartesian power and map

isaacy2012:
(Update of #162 with finalised name `cartesian_power` and `cartesian_power_map`)

Attempts to partially address #138 regarding the `cartesian_power`, `cartesian_power_map` functionalities. 

Implementation centers around replacing `std::get<I>(self.bases_)` in `cartesian_product` with `self.base_` for `cartesian_power`.

Moves common logic across `cartesian_product`, `cartesian_power` and their `map` variants to a common base trait class `cartesian_base` with enum selectors for [product/power] and [tuple/map].


tcbrindle:
Hi @isaacy2012, it looks like this PR is using the "repeat" names, did you intend to change them back to "power"? :)

isaacy2012:
> Hi @isaacy2012, it looks like this PR is using the "repeat" names, did you intend to change them back to "power"? :)

Oops sorry, forgot to push the revert :)

include/flux/op/cartesian_power_map.hpp
```diff
@@ -0,0 +1,65 @@
...
+    [[nodiscard]]
+    constexpr auto operator()(Func func, Seq&& seq) const
+    {
+        if constexpr(PowN == 0) {
+            return empty<std::invoke_result_t<Func>>;
+        } else {
+            return cartesian_power_map_adaptor<PowN, Func, std::decay_t<Seq>>(
+                std::move(func), FLUX_FWD(seq));
+        }
+    }
```
tcbrindle, line 53: The usual pattern we use for adaptor functions is that the base sequence is the first argument, followed by adaptor-specific arguments... e.g. filter(base, pred). The exception is for adaptors which take a variadic number of sequences, like `cartesian_product` and `zip`, where the language rules say the variadic pack has to go last in the argument list. (Possibly it would have been better to always pass the base sequence as the last argument for all adaptors to avoid this inconsistency, but that ship has sailed...)

I can see the intention here is to be consistent with `cartesian_product_map`, but on balance I think it's probably better to be consistent with the majority of adaptors and flip the order of the arguments here.

(Also, I should have noticed this earlier, sorry!)

```suggestion
    template <adaptable_sequence Seq, typename Func>
        requires multipass_sequence<Seq> &&
        detail::repeated_invocable<Func&, element_t<Seq>, PowN>
    [[nodiscard]]
    constexpr auto operator()(Seq&& seq, Func func) const
    {
        if constexpr(PowN == 0) {
            return empty<std::invoke_result_t<Func>>;
        } else {
            return cartesian_power_map_adaptor<PowN, Func, std::decay_t<Seq>>(
                std::move(func), FLUX_FWD(seq));
        }
    }
```

test/CMakeLists.txt
```diff
@@ -25,7 +25,9 @@ add_executable(test-flux
     test_bounds_checked.cpp
     test_cache_last.cpp
     test_cartesian_product.cpp
-    test_cartesian_product_with.cpp
+    test_cartesian_product_map.cpp
+    test_cartesian_power.cpp
+    test_cartesian_power_map.cpp
```
tcbrindle, line 30: Super pedantic: these should be in alphabetical order (so `test_cartesian_power.cpp` and `test_cartesian_power_map.cpp` before `test_cartesian_product.cpp`)

tcbrindle (changes requested):

tcbrindle:
Other than one change that I really should have noticed earlier (sorry!) and a tiny super-pedantic one, I think this is done! 

The only thing left would be some (placeholder) documentation in `docs/reference/adaptors.rst` now that the name is finalised, but I can do that if you like because Sphinx can be really fussy about things.

tcbrindle (approved):

tcbrindle:
@isaacy2012 are you happy for me to merge this?

isaacy2012:
> @isaacy2012 are you happy for me to merge this?

Yep, sounds good. Thanks for all your help and feedback :)

tcbrindle:
> > @isaacy2012 are you happy for me to merge this?
> 
> Yep, sounds good. Thanks for all your help and feedback :)

Thank you for working on it! It's very much appreciated

PR #168, opened 2024-02-01
Avoid overflow in cartesian_product size() impl

### LINKED ISSUE #167: cartesian_product size() can overflow
As noted when @isaacy2012 was adding `cartesian_power`, the current implementation of `size()` for `cartesian_product[_map]` uses built-in multiplication in a fold expression, and could easily overflow:

https://github.com/tcbrindle/flux/blob/85b016eee0893b4971faaa3dbb415d308212a276/include/flux/op/cartesian_base.hpp#L219-L227

It should use `num::checked_mul()` instead to guard against UB.

tcbrindle:
Fixes #167

### DIFF
--- a/include/flux/op/cartesian_base.hpp
+++ b/include/flux/op/cartesian_base.hpp
@@ -222,7 +222,9 @@ public:
                   && (sized_sequence<Bases> && ...))
     {
         return std::apply([](auto&... base) {
-            return (flux::size(base) * ...);
+            distance_t sz = 1;
+            ((sz = flux::num::checked_mul(sz, flux::size(base))), ...);
+            return sz;
         }, self.bases_);
     }
 
--- a/include/flux/op/cartesian_base.hpp
+++ b/include/flux/op/cartesian_base.hpp
@@ -221,9 +221,9 @@ public:
         requires (CartesianKind == cartesian_kind::product
                   && (sized_sequence<Bases> && ...))
     {
-        return std::apply([](auto&... base) {
-            distance_t sz = 1;
-            ((sz = flux::num::checked_mul(sz, flux::size(base))), ...);
+        return std::apply([](auto& base0, auto&... bases) {
+            distance_t sz = flux::size(base0);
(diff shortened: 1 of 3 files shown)


include/flux/op/cartesian_base.hpp
```diff
@@ -222,7 +222,9 @@ struct cartesian_traits_base_impl {
                   && (sized_sequence<Bases> && ...))
     {
         return std::apply([](auto&... base) {
-            return (flux::size(base) * ...);
+            distance_t sz = 1;
+            ((sz = flux::num::checked_mul(sz, flux::size(base))), ...);
```
isaacy2012, line 226: Is `flux::num` needed instead of just `num::`? It seems most places in the codebase use just `num::`.

include/flux/op/cartesian_base.hpp
```diff
@@ -222,7 +222,9 @@ struct cartesian_traits_base_impl {
                   && (sized_sequence<Bases> && ...))
     {
         return std::apply([](auto&... base) {
-            return (flux::size(base) * ...);
+            distance_t sz = 1;
```
isaacy2012, line 225: Though it should already be enforced by the public interface for `cartesian_product` and such, I wonder if there is value in static asserting `sizeof...(Bases) > 0` somewhere either in this function or in this struct given this default return value of 1?

include/flux/op/cartesian_base.hpp
tcbrindle: Yep, well spotted, this could just be `num::checked_mul`

include/flux/op/cartesian_base.hpp
tcbrindle: I've now changed the signature of the lambda to `[](auto& base0, auto&... bases)` which means we don't need to initialise `sz = 1`, and we get a check that the `self.base_` tuple has at least one element, the best of both worlds :)

isaacy2012 (approved):

tcbrindle:
Thanks for the review @isaacy2012 👍

PR #170, opened 2024-02-11
fix: documentation

goto40:
Great CppCon talk!

I found a minor bug in the documentation: the `sum(5)` yields 40 not 12 - I fixed this and added some comments.
Maybe you already fixed this somewhere - in this case just discard this change.

Somehow my auto-formatter also added some spaces (I can rework the PR if you prefer the original formatting).


tcbrindle:
Oh dear, that's a bit embarrassing for the first example people see! Thanks for spotting it.

Issue #172, opened 2024-02-14
Add zip_map adaptor


tcbrindle:
The element type of `flux::cartesian_product`, `cartesian_power` and `adjacent` is a tuple of the element type(s) of the underlying sequences. These three adaptors also provide `*_map` versions, which take an n-ary function and call it directly with the underlying elements. This is to avoid forming a tuple and then immediately destructuring it, as would happen in a subsequent call to `map(unpack(func))`.

The odd one out is `zip`, which doesn't provide a `_map` form. We should fix that.

tcbrindle:
Fixed by #184. Thanks @isaacy2012

PR #184, opened 2024-04-13
Add zip_map adaptor

isaacy2012:
Addresses, #172: The element type of flux::cartesian_product, cartesian_power and adjacent is a tuple of the element type(s) of the underlying sequences. These three adaptors also provide  *_map versions, which take an n-ary function and call it directly with the underlying  elements. This is to avoid forming a tuple and then immediately destructuring it, as would happen in a subsequent call to map(unpack(func)).

This commit adds the *_map version for the zip adaptor.


tcbrindle:
So first of all, please accept my apologies @isaacy2012 for letting this PR languish for months. I'm afraid I was pretty busy at the time you originally submitted it so I intended to review it later, and then I somehow I never got back to looking at it. I feel very bad!

The good news is that this PR looks great! I've just got a couple of small comments but otherwise this looks awesome.

If you've moved on in the intervening time and don't feel like coming back to this then I quite understand :). I'll be happy to do a final polish if you're busy with other things.

Thanks, and apologies again

include/flux/op/zip.hpp
```diff
@@ -169,6 +127,99 @@ struct sequence_traits<detail::zip_adaptor<Bases...>>
...
+
+public:
+    constexpr explicit zip_map_adaptor(Func&& func, decays_to<Bases> auto&&... bases)
+        : bases_(FLUX_FWD(bases)...), func_(std::move(func))
+    {}
+};
+
+struct zip_map_fn {
+    template <typename Func, adaptable_sequence... Seqs>
+        requires std::is_invocable_v<Func, element_t<Seqs>...>
```
tcbrindle, line 179: Minor: this constraint should use `Func&`, as we invoke the function object as an lvalue

```suggestion
        requires std::is_invocable_v<Func&, element_t<Seqs>...>
```

include/flux/op/zip.hpp
```diff
@@ -169,6 +127,99 @@ struct sequence_traits<detail::zip_adaptor<Bases...>>
...
+            return zip_adaptor<std::decay_t<Seqs>...>(FLUX_FWD(seqs)...);
+        }
+    }
+};
+
+template <typename Func, sequence... Bases>
+struct zip_map_adaptor : inline_sequence_base<zip_map_adaptor<Func, Bases...>> {
+private:
+    pair_or_tuple_t<Bases...> bases_;
+    Func func_;
```
tcbrindle, line 166: Minor: the `func_` member could be marked with `FLUX_NO_UNIQUE_ADDRESS` (although I'm pretty inconsistent with this myself)

```suggestion
    FLUX_NO_UNIQUE_ADDRESS Func func_;
```

include/flux/op/zip.hpp
```diff
@@ -169,6 +127,99 @@ struct sequence_traits<detail::zip_adaptor<Bases...>>
...
+};
+
+struct zip_map_fn {
+    template <typename Func, adaptable_sequence... Seqs>
+        requires std::is_invocable_v<Func, element_t<Seqs>...>
+    [[nodiscard]]
+    constexpr auto operator()(Func func, Seqs&&... seqs) const
+    {
+        if constexpr (sizeof...(Seqs) == 0) {
+            return empty<std::invoke_result_t<Func>>;
```
tcbrindle, line 184: At the moment, `flux::empty<T>` is constrained to require `std::is_object_v<T>`, so this will be a problem if `func()` returns a reference type. 

I have to confess that I can't remember the reason for having that constraint on `empty`! It doesn't seem like it would be problematic to remove it, so that we can make this work.

test/test_zip_map.cpp
tcbrindle, line 1: The tests are great, but it would also be nice to have a test for `zip_map` with one sequence (to verify it does the same as `map`), and also for zero sequences just to check that the empty case works.

tcbrindle (changes requested):

isaacy2012:
Hi @tcbrindle, no worries at all, I'll address those changes this weekend :)

test/test_zip_map.cpp
```diff
@@ -0,0 +1,274 @@
...
+        static_assert(flux::bidirectional_sequence<Z const>);
+        static_assert(flux::random_access_sequence<Z const>);
+        static_assert(flux::sized_sequence<Z const>);
+        static_assert(flux::bounded_sequence<Z const>);
+
+        static_assert(std::same_as<flux::element_t<Z>, int>);
+        static_assert(std::same_as<flux::value_t<Z>, int>);
+        static_assert(std::same_as<flux::rvalue_element_t<Z>, int>);
+
+        // Believe it or not these are correct -- we've zipped two ref adaptors
```
tcbrindle, line 47: No change required, but this comment has slipped past copy-and-paste -- there's nothing surprising about the following lines with `zip_map`

tcbrindle (approved):
Looks good to me. Great work! It would be nice to have some (at least placeholder) documentation but I can add this after merging.

Please merge the latest Flux changes from the main branch so we can have another CI run to make sure there are no problems, then this will be good to go.

tcbrindle (approved):

tcbrindle:
Thanks @isaacy2012!

Issue #191, opened 2024-06-29
Add filter_map adaptor


Guekka:
Rust [has it](https://doc.rust-lang.org/nightly/std/iter/struct.FilterMap.html)

It avoids code like this
```cpp
        .map([&files_in_directory](const btu::Path &path) -> std::optional<btu::Path> {
            if (const auto it = files_in_directory.find(path); it != files_in_directory.end())
                return it->second;
            return {};
        })
        .filter([](const auto &path) { return path.has_value(); })
        .map([](const auto &path) { return path.value(); })
```

Please correct me if I missed an equivalent

tcbrindle:
So this would be an adaptor that takes a function of the form `(T) -> optional<U>`, and then yields those elements for which the optional is engaged, right?

I definitely think that that would be a nice thing to have.

We could re-use the `optional_like` concept for testing the return type of the mapping function:

```cpp
template <typename O>
concept optional_like =
    std::default_initializable<O> &&
    std::movable<O> &&
    requires (O& o) {
        { static_cast<bool>(o) };
        { *o } -> flux::detail::can_reference;
    };
```

This would allow not just `flux::optional` and `std::optional` to be used, but also (for better or worse) raw pointers and smart pointers.

We could also think about adding a convenience function which does `filter_map(std::identity)` for the case where the sequence elements are already optional-like -- something like `filter_deref`?

Guekka:
> So this would be an adaptor that takes a function of the form (T) -> optional<U>, and then yields those elements for which the optional is engaged, right?

Indeed. Sorry for not describing it, I assumed it was known to you.

I've given some thought to your idea about `optional_like`. I don't like the idea of dereferencing pointers, as it could lead to dangerous behavior with smart pointers. Imagine this case ([godbolt](https://godbolt.org/z/Gd8v1osGf))
```cpp
#include <vector>
#include <memory>
#include <iostream>

#include "https://raw.githubusercontent.com/tcbrindle/flux/main/single_include/flux.hpp"

// unique_ptr has compile error in map() and no time to investigate
// doesn't map() pass arguments as rvalue?
template <class T>
using Ptr = std::shared_ptr<T>;

int main() {
    std::vector<Ptr<int>> vec(100);

    flux::from(std::move(vec))
        .map([] (Ptr<int> ptr) { return *ptr; }) // oops
        // memory is freed...
        .for_each([](auto val) { std::cout << val; });
}
```

[...]

PR #192, opened 2024-07-05
implement filter_map

Guekka:
Adresses #191
The functionality is there, but I am still trying to understand why the `static_assert` fail


tcbrindle:
Thanks for working on this @Guekka.

> The functionality is there, but I am still trying to understand why the static_assert fail

The function `is_even_opt` in `test_filter_map.cpp` returns a prvalue `std::optional<int>`:

https://github.com/Guekka/flux/blob/95decaf4aeb18506804f858129b3989f32bebcab/test/test_filter_map.cpp#L15

When reading from the pipeline, the temporary optional gets dereferenced in this lambda:

https://github.com/Guekka/flux/blob/95decaf4aeb18506804f858129b3989f32bebcab/include/flux/op/filter_map.hpp#L26

This returns a reference to the temporary optional's internal `int`. Unfortunately, when the `read_at()` call returns, the temporary `optional` gets destroyed, and the returned reference is immediately dangling. (This took me a little while to figure out, but I'm very pleased that the constexpr testing picked it up!)

One solution is to try to detect the problem, by noticing when dereffing the optional-like would return an rvalue reference, and returning by value instead. Like so:

```cpp
struct filter_map_fn {
    // If dereffing the optional would give us an rvalue reference,
    // prevent a probable dangling reference by returning by value instead
    template <typename T>
    using strip_rvalue_ref_t = std::conditional_t<
        std::is_rvalue_reference_v<T>, std::remove_reference_t<T>, T>;

   template <adaptable_sequence Seq, typename Func>
        requires (std::invocable<Func&, element_t<Seq>> &&
                  optional_like<std::remove_cvref_t<std::invoke_result_t<Func&, element_t<Seq>>>>)
    constexpr auto operator()(Seq&& seq, Func func) const
    {
        return flux::map(FLUX_FWD(seq), std::move(func))
            .filter([](auto&& opt) { return static_cast<bool>(opt); })
            .map([](auto&& opt) -> strip_rvalue_ref_t<decltype(*FLUX_FWD(opt))> {
                return *FLUX_FWD(opt);
            });
    }
};
```

Making this change fixes the original lifetime problem. Unfortunately this `static_assert` still fails because some `new`s later in the test are never `delete`d, but this is pretty easy to fix :)

Hope this helps!

include/flux/op/filter_map.hpp
```diff
@@ -0,0 +1,63 @@
...
+            template <adaptable_sequence Seq, typename Func>
+                requires (std::invocable<Func&, element_t<Seq>> &&
+                          optional_like<std::remove_cvref_t<std::invoke_result_t<Func&, element_t<Seq>>>>)
+            constexpr auto operator()(Seq&& seq, Func func) const
+            {
+                return flux::map(FLUX_FWD(seq), std::move(func))
+                    .filter([](auto&& opt) { return static_cast<bool>(opt); })
+                    .map([](auto&& opt) -> decltype(auto) { return *FLUX_FWD(opt); });
+            }
+        };
```
tcbrindle, line 28:
```suggestion
        struct filter_map_fn {
            // If dereffing the optional would give us an rvalue reference,
            // prevent a probable dangling reference by returning by value instead
            template <typename T>
            using strip_rvalue_ref_t = std::conditional_t<
                std::is_rvalue_reference_v<T>, std::remove_reference_t<T>, T>;

            template <adaptable_sequence Seq, typename Func>
                requires (std::invocable<Func&, element_t<Seq>> &&
                          optional_like<std::remove_cvref_t<std::invoke_result_t<Func&, element_t<Seq>>>>)
            constexpr auto operator()(Seq&& seq, Func func) const
            {
                return flux::map(FLUX_FWD(seq), std::move(func))
                    .filter([](auto&& opt) { return static_cast<bool>(opt); })
                    .map([](auto&& opt) -> strip_rvalue_ref_t<decltype(*FLUX_FWD(opt))> {
                        return *FLUX_FWD(opt);
                    });
            }
        };
```

Guekka:
Thank you @tcbrindle. Debugging why something is not constexpr is complex

test/test_filter_map.cpp
```diff
@@ -0,0 +1,164 @@
+
+// Copyright (c) 2024 Tristan Brindle (tcbrindle at gmail dot com)
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+#include "catch.hpp"
+
+#include <array>
+#include <utility>
```
tcbrindle, line 9: Looking at the [CI failure](https://github.com/tcbrindle/flux/actions/runs/9971483316/job/27554724376#step:6:308), it seems like you need to `#include <optional>` here

include/flux/core/inline_sequence_base.hpp
```diff
@@ -7,6 +7,7 @@
 #define FLUX_CORE_INLINE_SEQUENCE_BASE_HPP_INCLUDED
 
 #include <flux/core/sequence_access.hpp>
+#include <flux/core/simple_sequence_base.hpp>
```
tcbrindle, line 10: If we're going to re-use it in other places, I'd prefer to move the `optional_like` concept into `core/concepts.hpp` (but keep it in a `detail` namespace for now) and remove this `#include`

include/flux/core/inline_sequence_base.hpp
```diff
@@ -271,6 +272,15 @@ struct inline_sequence_base {
     [[nodiscard]]
     constexpr auto filter(Pred pred) &&;
 
+    template <typename Func>
+    requires (std::invocable<Func&, element_t<Derived>> &&
+        detail::optional_like<std::invoke_result_t<Func&, element_t<Derived>>>)
```
tcbrindle, line 277: Formatting: indent the requires clause

```suggestion
        requires (std::invocable<Func&, element_t<Derived>> &&
            detail::optional_like<std::invoke_result_t<Func&, element_t<Derived>>>)
```

include/flux/op/filter_map.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+#ifndef FLUX_OP_FILTER_MAP_HPP_INCLUDED
+#define FLUX_OP_FILTER_MAP_HPP_INCLUDED
+
+#include <flux/core.hpp>
+#include <flux/op/map.hpp>
+
+namespace flux {
+
+    namespace detail {
```
tcbrindle, line 14: Formatting: nested namespaces don't add a level of indentation

(I'll get round to adding a .clang-format config eventually...)

include/flux/op/filter_map.hpp
```diff
@@ -0,0 +1,68 @@
...
+        struct filter_map_fn {
+            // If dereffing the optional would give us an rvalue reference,
+            // prevent a probable dangling reference by returning by value instead
+            template <typename T>
+            using strip_rvalue_ref_t = std::conditional_t<
+                std::is_rvalue_reference_v<T>, std::remove_reference_t<T>, T>;
+            template <adaptable_sequence Seq, typename Func>
+                requires (std::invocable<Func&, element_t<Seq>> &&
+                          optional_like<std::remove_cvref_t<std::invoke_result_t<Func&, element_t<Seq>>>>)
+            constexpr auto operator()(Seq&& seq, Func func) const
```
tcbrindle, line 25: Formatting: add a blank line between the using declaration and the following declaration for readability

```suggestion
            template <typename T>
            using strip_rvalue_ref_t = std::conditional_t<
                std::is_rvalue_reference_v<T>, std::remove_reference_t<T>, T>;
                
            template <adaptable_sequence Seq, typename Func>
                requires (std::invocable<Func&, element_t<Seq>> &&
                          optional_like<std::remove_cvref_t<std::invoke_result_t<Func&, element_t<Seq>>>>)
            constexpr auto operator()(Seq&& seq, Func func) const
```

test/test_filter_map.cpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    // A predicate that always returns true returns what it was given
+    {
+        int arr[] = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
+        auto filtered = flux::filter_map(flux::ref(arr), [](auto&& i) { return std::optional{i}; });
+
+        if (!check_equal(arr, filtered)) {
+            return false;
+        }
```
tcbrindle, line 67: Please use the `STATIC_CHECK` macro rather than a manual `if` statement as it gives better diagnostics if there's a constexpr test failure (here and in other places in this file)

```suggestion
        STATIC_CHECK(check_equal(arr, filtered));
```

(I guess you copied this from `test_filter.cpp`, which was mostly written before the macro existed and should probably be updated to use it)

Guekka:
Cannot reproduce the CI failure locally on Windows. Will try later on Linux

tcbrindle (changes requested):
This is looking good. I've suggested a few changes but they're pretty minor.

I the long term, I'd prefer `filter_map` to have its own dedicated adaptor class rather than using a chain of three existing adaptors, but doing it this way is fine for now.

We also need some documentation (or at least placeholders) for `filter_map` and `filter_deref` in `docs/reference/adaptors.rst` to let people know that they exist. If this is tricky then let me know and I can add it to the PR before merging.

tcbrindle:
> Cannot reproduce the CI failure locally on Windows. Will try later on Linux

Yeah, it seems like it only happens with libstdc++ -- both MSVC and libc++ pull in `std::optional` from somewhere else

Guekka:
Resolved reviews. Still have to write doc

Guekka:
Please review thoroughly the docs @tcbrindle, I'm not entirely sure of what I wrote
Also, thank you for sticking with me, it would probably have been faster for you to write it all

I will rebase the PR once you approve it

docs/reference/adaptors.rst
```diff
@@ -814,6 +814,86 @@ You can pass a reference to a sequence into an adaptor using :func:`flux::ref` o
...
+
+        * :func:`flux::filter_map`
+
+``filter_map``
+^^^^^^^^^^^^^^
+
+..  function::
+    template <sequence Seq, typename Func> \
+        requires std::invocable<Func&, element_t<Seq>> && \
+                 optional_like<std::invoke_result_t<Func&, element_t<Derived>>> \
```
tcbrindle, line 864:
```suggestion
                 optional_like<std::invoke_result_t<Func&, element_t<Seq>>> \
```

docs/reference/adaptors.rst
```diff
@@ -814,6 +814,86 @@ You can pass a reference to a sequence into an adaptor using :func:`flux::ref` o
...
+    auto filter_map(Seq seq, Func func) -> sequence auto;
+
+    Has the behaviour of both :func:`flux::filter` and :func:`flux::map`.
+    The returned sequence contains only the values for which :var:`func` returns a value, and the values themselves are the result of applying :var:`func` to the original values.
```
tcbrindle, line 868: Again, I'd say something like:

```suggestion
    Performs both filtering and mapping using a single function.

    Given a unary function :var:`func` returning an "optional-like" type, the returned adaptor filters out those elements for which :var:`func` returns a "disengaged" optional (that is, those which return :texpr:`false` after conversion to :texpr:`bool`). It then dereferences the remaining elements using :expr:`operator*`.

    Equivalent to::

        map(seq, func)
           .filter([](auto&& arg) { return static_cast<bool>(arg) })
           .map([](auto&& arg) -> decltype(auto) { return *std::forward(arg); });
```

tcbrindle (changes requested):
Good job on the documentation, I've made some suggestions to add a bit more description but otherwise it's good. I like making the doc target watch all the .rst files 👍 

There's one change I don't understand with `FLUX_NO_UNIQUE_ADDRESS`, if this wasn't intentional then please revert it (or else let me know why it's needed)

Other than that it's just some minor formatting things

include/flux/macros.hpp
```diff
@@ -18,7 +18,7 @@
 #define FLUX_ALWAYS_INLINE
 #endif
 
-#define FLUX_NO_UNIQUE_ADDRESS [[no_unique_address]]
+#define FLUX_NO_UNIQUE_ADDRESS
```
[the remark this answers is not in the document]
Guekka: Good catch, it was only for development. The semi proper fix would be using `[[msvc:no_unique_address]]` on Windows (see [this issue](https://github.com/llvm/llvm-project/issues/49358))
For now, I will remove this change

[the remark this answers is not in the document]
tcbrindle: Yeah, supporting `[[msvc::no_unique_address]]` was the reason for making it a macro in the first place, but then I read something about MSVC mis-compiling things when using this attribute so I never actually implemented it. But that was a long time ago and it's probably fixed now, so it's probably worth doing at some point.

Guekka:
Fixed. A clang format file would really be helpful, even if it means losing some custom formatting, I think it would be worth it for contributors

tcbrindle (approved):

tcbrindle:
Thanks for your work on this @Guekka!

PR #234, opened 2025-04-14
Fix CI module builds on Linux

tcbrindle:
The  various module builds have broken on Linux Clang (still working on MacOS) for reasons I don't understand. Let's try re-enabling them one at a time to figure it out.


tcbrindle:
Looks like Clang 17 module builds on Linux will remain broken (though they work on MacOS). No idea what's changed...

PR #236, opened 2025-04-14
Try enabling module builds on Windows

tcbrindle:
Just to see what happens


tcbrindle:
Still not there with modules on Windows, although we're now getting a "sorry, not yet implemented" ICE rather than the previous "the impossible happened", so I guess that's progress?

PR #250, opened 2025-09-02
Periodic test of MSVC modules support

tcbrindle:
Since I've been mouthing off on Reddit about it not working


tcbrindle:
Not yet

